PR #169, opened 2026-07-28
Resolve Android back to the route hierarchy, not browser history

PeterXMR:
Rebased onto `main` after #170 landed. That PR added the Android back-button bridge independently, so **this no longer touches the native shell at all** — `MainActivity.java` and `nativeBridge.ts` are unchanged. The `linky-native-back-button` cancelable-event mechanism from #170 is kept as-is; only the web-side resolution changes.

## Why `history.back()` doesn't work here

`navigateTo()` pushes a new hash entry for every navigation — **including the top-left back button** (all 39 call sites use `location.assign`, none use `replace`). So history only ever grows, and stepping back through it replays visited pages instead of moving up the hierarchy.

Both consequences reproduced on `Pixel_6_API_36_Linky` against #170 as merged:

**Bounce.** On `#settings/units`, tapping the top-left button goes to `#settings` (history 5 → 6). Hardware back then returns to `#settings/units` — the page just left.

```
at: #settings/units (history.length=5)
tapped-arrow -> now at #settings (history.length=6)
hardware back -> #settings/units      <- bounce
```

**Dead button.** Landing on `#chat/<id>` with `history.length == 1`, as a notification or deep-link cold start does, `preventDefault()` is called but `history.back()` has nowhere to go. Back neither navigates nor exits:

```
at: #chat/unknown:...ab (history.length=1)
hardware back -> hash unchanged, foreground=true    <- dead
```

With this branch both resolve correctly (`-> #contacts` in each case).

## What changed

Back resolves through `resolveBackAction()` — the same mapping the top-left button uses — so the two cannot drift as routes are added. Its `switch` is exhaustive over `Route["kind"]`, so a new route with no back target fails typecheck rather than silently exiting the app. The bank payment offer keeps its `×` glyph and still marks the offer minimized via `returnFromBankPaymentOffer()`, matching #170.

Open overlays are dismissed before the route moves. `AuthenticatedLayout` renders its modals as siblings of the route content, so nothing unmounts them on navigation; without this, back moved the route out from under an open payment confirmation and left its pending promise unresolved. `resolveNativeBackAction()` orders the surfaces (topmost modal → scan overlay → menu → parent route) and is also what decides whether to claim the press, so the two can never disagree.

On a root screen nothing is claimed, the event stays uncancelled, and the shell closes the app exactly as it does today.

`topbarConfig.ts` is net −292 lines: 35 near-identical `{icon, label, onClick}` blocks collapse into one switch.

## Verification

Real debug build on the emulator, injecting actual `KEYCODE_BACK` presses and reading state over CDP:

- 18 navigation assertions across settings / wallet / topup / mints / relays / evolu-server / profile / transactions chains
- `#settings` returns to whichever root the user came from (`#wallet` vs `#contacts`)
- chat thread → `#contacts`
- back at a root screen exits the app
- scan overlay is dismissed without changing the route, and the next press then leaves the page

Plus 16 unit tests: `resolveBackAction` (every non-root route has a target, both roots return `null`, bank-offer minimize side effect) and `resolveNativeBackAction` ordering.

Pre-existing unrelated failures in the suite (`useInboxNotificationsSync`, `storage.preferences`, `cashuOwnerLane`, and the Playwright specs vitest picks up) are unchanged by this branch.

## Known remaining wart

While the QR scan overlay is open the first two back presses are swallowed before reaching any handler — no native-scanner logs, so it looks like the WebView's fullscreen camera layer consuming them. The third closes the scanner without changing route. This is not introduced here and behaves the same on `main`; I left it rather than reach into Capacitor's `WebChromeClient`.


kaladivo:
Thanks @PeterXMR! This branch had conflicted with `main` again after 9559a90 landed, so I've rebased it and reopened it as #196 — your commit is preserved there with you as the author. Closing this one in favor of #196.

PR #201, opened 2026-07-31
Implement ready-for-pickup batch: app-shell, Cashu, and typing refactors (#178-#188)

### LINKED ISSUE #178: Mount the full app-shell composition only when authenticated
**Context**

This is a remaining, deliberately deferred refactor from the July 2026 full-repo code audit. Most findings from that audit were addressed in PR #171; this one was left for later.

**Problem**

In `apps/web-app/src/app/AppShell.tsx`, `useAppShellComposition()` (the full authenticated hook composition — domains, effects, subscriptions) always runs, even when the user is unauthenticated and only the onboarding screens are shown. This wastes work on the onboarding path and tightly couples the onboarding UI to the entire authenticated shell.

**Suggested approach**

Split a small, minimal unauthenticated/onboarding composition (language selection, welcome/restore/create-account flows) out from the full authenticated composition, and only mount the full composition once the user is actually logged in. This gating should happen at the `AppShell` level (e.g. choosing which composition hook to call based on auth state at the top of the tree) rather than conditionally invoking hooks within a single component, to keep hook call order stable.

### LINKED ISSUE #179: Replace the ~120-parameter route-prop transport with per-domain contexts/hooks
## Context

This is a remaining, deliberately deferred refactor from the July 2026 code audit. Most audit findings were addressed in PR #171 (route bundles are now memoized), but the underlying transport shape in `apps/web-app/src/app/routes/useSystemRouteProps.ts` was left unchanged.

## Problem

Route pages receive their props through a large flat forwarding layer: `useSystemRouteProps` alone declares roughly 120 individually-typed parameters, and `buildPeopleRouteProps` / `buildMoneyRouteProps` (in `apps/web-app/src/app/routes/props/`) do the same kind of flat forwarding for their respective route groups. This is brittle — adding or renaming a single piece of state means touching several files in lockstep — and it drowns out the actual data flow between shell state and the pages that consume it.

## Suggested approach

At a high level, replace the flat prop bundles with per-domain route hooks or narrow contexts that pages read directly, rather than threading everything through a central prop-assembly layer. The typed context consumer hooks already in `apps/web-app/src/app/context/AppShellContexts.tsx` (`useAppShellCore`, `useAppShellActions`, `useAppShellRouteContext`) point the way. This should be migrated incrementally, one route group per PR, rather than as a single large rewrite.

### LINKED ISSUE #180: Type persistence boundaries with real Evolu query result types instead of *RowLike aliases
## Context

This is a remaining, deliberately deferred refactor identified in the July 2026 full-repo code audit. Most audit findings were addressed in PR #171 (audit-fixes), but this one was left for a follow-up.

## Problem

`apps/web-app/src/app/types/appTypes.ts` (around line 119) defines loose `OptionalText` / `OptionalNumber` unions (`string | number | boolean | bigint | symbol | { toString(): string } | null | undefined`, etc.) and, elsewhere in the app, hand-rolled `*RowLike` types are used to describe data coming out of Evolu queries. These stand in for the real Evolu row shapes and hide the actual data contracts at persistence boundaries — a value typed this loosely tells callers almost nothing about what a given column can actually contain, and drifts silently from the real schema over time.

## Suggested approach

At a high level, derive these types from the Evolu schema / query result types instead of redefining them by hand — Evolu already exports typed row/query result types for a given schema and query. Threading those real types through the hooks and components that currently consume the loose aliases would remove the need for `OptionalText`/`OptionalNumber`/`*RowLike` entirely for those cases. This is likely best done incrementally, table by table, rather than as one large sweep.

### LINKED ISSUE #181: Unify the two kind-1059 inbox processors into one decoded-event pipeline
## Context

This is a remaining, deliberately deferred refactor from the July 2026 code audit. Most audit findings were fixed in PR #171, and that round already shared the identity resolver between the two inbox code paths via `apps/web-app/src/app/hooks/messages/chatNostrProtocol.ts`. This item was explicitly called out as a dedicated protocol refactor rather than a quick fix.

## Problem

There are two separate kind-1059 gift-wrap inbox processors:

- `useChatNostrSyncEffect` (`apps/web-app/src/app/hooks/messages/useChatNostrSyncEffect.ts`) — active-chat sync
- `useInboxNotificationsSync` (`apps/web-app/src/app/hooks/messages/useInboxNotificationsSync.ts`) — global inbox sync

Both independently unwrap, validate, and persist gift-wrapped events (`processWrap` in each), and both subscribe to relays and select which relay set to use. Because the decode/validate/persist logic is duplicated rather than shared, the two paths have already drifted — notably in how reaction `k` tags are handled and in relay selection — and are likely to keep drifting as either path is touched independently.

## Suggested approach

Extract a single decoded-event pipeline (unwrap → validate inner rumor → classify message/reaction/edit/payment-notice → persist/notify) that both `useChatNostrSyncEffect` and `useInboxNotificationsSync` feed into, with the differences between the two entry points (which chats are in scope, notification behavior) passed in as parameters rather than reimplemented. This should land as its own protocol refactor, with tests covering the existing gift-wrap rejection rules (outer/tagged pubkey checks, nested NIP-44 payload rejection, malformed inner event checks) to guard against regressions during the merge.

### LINKED ISSUE #182: Cashu refactors deferred from the audit: accept/persist, melt executor, proof state, restore scanner
**Context**

These are remaining refactor items from the July 2026 code audit of `apps/web-app/src/app/hooks/cashu/` (and `useLightningPaymentsDomain`). Most audit findings were already fixed in PR #171, but the items below were deliberately deferred as too risky for an automated pass since they touch money-moving code paths.

**Problem**

- [ ] Shared accept-and-persist service (`useNpubCashClaim`)
- [ ] Shared melt executor (`useLightningPaymentsDomain`)
- [ ] Proof-state primitive (`useCashuTokenChecks`)
- [ ] Restore scanner split (`useRestoreMissingTokens`)

Each of these hooks has grown to own multiple concerns that are duplicated or overgrown across the cashu domain, making them harder to reason about and test in isolation.

**Suggested approach**

Tackle one item per PR rather than combining them. For each: extract a pure service/helper first, re-point the existing hook at it without changing behavior, and lock the current behavior with tests before moving any logic — these paths move money, so behavior-preserving steps first, structural cleanup second.

### LINKED ISSUE #183: Split the 660-line usePayContactWithCashuMessage payment workflow into testable steps
## Context

This is a remaining, deliberately deferred refactor from the July 2026 code audit. Most audit findings for this codebase were already fixed in PR #171; this one was left for a follow-up pass since it's a larger structural change rather than a bug fix.

## Problem

`apps/web-app/src/app/hooks/payments/usePayContactWithCashuMessage.ts` is one large hook that mixes several concerns in a single flow: mint/token selection, Cashu token swap and payload building, retryable amount-fallback logic, Nostr message + payment notice publishing, and Evolu persistence/cleanup. That makes the hook hard to read, hard to unit test in isolation, and risky to change since a small edit to one concern (e.g. retry logic) can affect unrelated concerns (e.g. persistence).

## Suggested approach

Split the workflow into sequential pure/testable steps rather than one monolithic hook body, roughly:

1. select mint and tokens for the payment amount
2. build the Cashu swap/send payload
3. publish the chat message and payment notice over Nostr
4. persist the resulting transaction/token state and clean up

Extract each step into its own small, independently testable function, and keep the hook itself as a thin orchestrator that wires them together and manages React state/effects. Because this path moves money, add tests locking in current behavior around each extracted step before/while moving the logic out, so the refactor can't silently change payment behavior.

### LINKED ISSUE #184: Remove the no-op record*OwnerWrite callbacks threaded through ~30 call sites
## Context

This is a remaining cleanup item deliberately deferred from the July 2026 code audit. Most audit findings were fixed in PR #171. Owner rotation counters now derive deterministically from local Evolu history (`evolu_history`) plus synced `ownerMeta` rotation snapshots, rather than from per-browser localStorage edit counters. As a result, the `recordContactsOwnerWrite` / `recordCashuOwnerWrite` / `recordMessagesOwnerWrite` / `recordTransactionsOwnerWrite` callbacks defined in `apps/web-app/src/app/hooks/useEvoluContactsOwnerRotation.ts` became no-ops, but were left in place to keep that PR scoped.

## Problem

The four `record*OwnerWrite` callbacks do nothing anymore, but they are still threaded through as parameters/props/call-sites across the codebase (roughly 30+ call sites across ~5 files). This adds noise and unnecessary prop-drilling with no behavioral purpose.

## Suggested approach

- Delete the callback definitions at the source in `useEvoluContactsOwnerRotation.ts`.
- Mechanically remove every parameter, prop, and call-site that threads them through, in one PR — let the TypeScript compiler drive the sweep (typecheck will surface every remaining reference).
- No behavior change is expected.
- Verify with `bun run check-code` (typecheck → eslint --fix → prettier --write) before merging.

### LINKED ISSUE #185: Extract usePushRegistrationLifecycle out of useRelayDomain
**Context**

Remaining item from the July 2026 code audit. Most findings from that audit were addressed in PR #171, but this refactor was deliberately deferred.

**Problem**

`apps/web-app/src/app/hooks/useRelayDomain.ts` mixes two unrelated concerns. The first ~300 lines implement a self-contained push-registration subsystem (subscription setup, VAPID key validation, installation-id handling, re-registration on subscription change) hiding behind the "relay" name, even though this logic has nothing to do with relay list management.

**Suggested approach**

Extract a `usePushRegistrationLifecycle` hook with a narrow interface (enabled state, pubkeys, relay list in; registration status out), and have `useRelayDomain` consume it or sit beside it as a sibling hook. This should be a pure move with no behavior change, verified by `bun run check-code` and by exercising the existing push registration flows.

### LINKED ISSUE #186: Remove dead helper and unused parameters in topupMintClaim
## Context

Remaining loose end from the July 2026 code audit. Most findings from that audit were fixed in PR #171, but this one was skipped because a clean removal spans the helper itself, its test file, and the composition call site, which crossed file-ownership boundaries for that pass.

## Problem

`apps/web-app/src/app/hooks/topup/topupMintClaim.ts` exports `shouldKeepTopupQuoteAfterClaimError(error, isOutputsAlreadySignedError)`, but the function body voids both parameters and unconditionally returns `false`:

```ts
export const shouldKeepTopupQuoteAfterClaimError = (
  error: unknown,
  isOutputsAlreadySignedError: (error: unknown) => boolean,
): boolean => {
  void error;
  void isOutputsAlreadySignedError;
  // mintTopupProofs already runs the deterministic restore loop on OAS.
  // ...
  return false;
};
```

It's dead logic that always resolves the same branch, yet it is still imported and called from `apps/web-app/src/app/useAppShellComposition.tsx` (the `shouldKeepTopupQuoteAfterClaimError(...)` check around the mint-claim error handling), and has a dedicated test case in `apps/web-app/tests/topupMintClaim.test.ts`.

## Suggested approach

Remove the dead helper and its unused parameters, and update the two dependent spots in the same small PR:
- `apps/web-app/src/app/hooks/topup/topupMintClaim.ts` — delete `shouldKeepTopupQuoteAfterClaimError` (keep `isTopupMintQuoteClaimableState`, which is still used).
- `apps/web-app/src/app/useAppShellComposition.tsx` — drop the call site and inline whatever the always-`false` branch actually implied for the claim-error handling.
- `apps/web-app/tests/topupMintClaim.test.ts` — remove the corresponding test case and unused import.

Verify with `bun run check-code` and the topup vitest suite afterward.

### LINKED ISSUE #187: Fix pre-existing vitest failures and Playwright spec leakage into vitest
**Context**

These test failures pre-date the July 2026 audit and reproduce identically on `main`; this was verified while working on PR #171 (audit-fixes), so none of them were introduced by that branch.

**Problem**

- [ ] `cashuOwnerLane.test.ts` uses fake owner ids that fail Evolu `OwnerId` validation
- [ ] 5 failing toast assertions in `useInboxNotificationsSync.test.tsx`
- [ ] The `cashuWallet` fallback-keyset test fails
- [ ] `vitest.config.ts` has no `exclude`, so Playwright `.spec.ts` files under `tests/` get picked up and run by vitest

**Suggested approach**

- Start with the `vitest.config.ts` exclude — likely a one-line fix (exclude `tests/**` or `*.spec.ts` from vitest) and it should cut a lot of the surrounding noise.
- For `cashuOwnerLane.test.ts`, generate valid Evolu `OwnerId`s in the test fixtures instead of fake ids.
- For the toast assertions and the fallback-keyset test, re-examine each to determine whether the test expectations or the underlying code drifted, then update whichever is wrong.

### LINKED ISSUE #188: Resolve the 4 remaining exhaustive-deps warnings in useAppShellComposition
## Context

This is a remaining loose end from the July 2026 full-repo code audit. Most findings from that audit were addressed in PR #171, but a handful of non-blocking lint warnings in `apps/web-app/src/app/useAppShellComposition.tsx` were left unfixed.

## Problem

`bunx eslint src/app/useAppShellComposition.tsx` currently reports 4 `react-hooks/exhaustive-deps` warnings in that file:

- React Hook `React.useCallback` has unnecessary dependencies: `selectedChatContact.npub` and `selectedChatContact.unknownPubkeyHex`. Either exclude them or remove the dependency array.
- React Hook `React.useCallback` has a missing dependency: `setMintIconUrlByMint`. Either include it or remove the dependency array. (two separate callbacks report this)
- React Hook `React.useCallback` has a missing dependency: `getCashuTokenMessageInfo`. Either include it or remove the dependency array.

None of these currently fail the lint step (they're warnings, not errors), so they don't block `bun run check-code`, but they should still be resolved.

## Suggested approach

Each of the 4 warnings needs an individual judgment call rather than a blanket fix — this file has previously had real bugs caused by missing hook dependencies (e.g. the owner-rotation lane revert fixed in PR #171), so dependency arrays here shouldn't be adjusted mechanically. For each warning, consider:

- Adding the missing/flagged dependency if doing so is safe and doesn't change intended behavior.
- Restructuring with refs or a "latest value" pattern (e.g. a ref updated via effect, read inside the callback) when a value is intentionally meant to stay stale/current-without-retriggering the callback identity.
- Falling back to a line-scoped `eslint-disable-next-line react-hooks/exhaustive-deps` only when neither of the above applies, with a comment explaining why the dependency is intentionally omitted.

Composition hooks in `apps/web-app/src/app/hooks/composition/` follow similar patterns and may be worth checking for the same class of issue while this is being cleaned up.

jarvisonsmart:
## Summary

Batch of maintenance/refactor issues merged into one branch. `bun run check-code` passes and all 529 vitest unit tests (105 files, `@linky/web-app` + `@linky/core`) pass on the merged result.

> **Note on history:** this branch was rewritten into linear history — one commit per issue, zero merge commits — and force-pushed. The earlier merge-based history is gone on purpose; review the 9 commits as they stand now.

- **#186** — Removed the dead helper and unused parameters from the `topupMintClaim` flow.
- **#184** — Removed the no-op `record*OwnerWrite` callbacks that were threaded through ~30 call sites; rotation counters already derive from Evolu history, so the callbacks were pure noise.
- **#188** — The 4 remaining `react-hooks/exhaustive-deps` warnings in `useAppShellComposition` were already resolved on main by aa7c771; verified nothing remains.
- **#187** — Pre-existing vitest failures and Playwright spec leakage into vitest were already fixed on main by 8e81e0e (vitest wired into scripts and CI, Playwright `*.spec.ts` excluded); verified green here.
- **#178** — App-shell composition is now mounted only when authenticated, instead of running the full composition graph for logged-out users.
- **#183** — Split the ~660-line `usePayContactWithCashuMessage` payment workflow into separate testable steps (mint selection, melt, token swap/persist, message send), with unit tests over the extracted pieces.
- **#182** — All four items from the audit are implemented: the shared accept-and-persist service, the melt executor, the proof-state primitive, and the restore scanner. (The earlier version of this PR description understated this as "first increment only" — that was stale; see the clarification comment on the issue.)
- **#180** — Persistence boundaries are now typed with real Evolu query result row types instead of the hand-rolled `*RowLike` aliases.
- **#179** — Replaced the ~120-parameter route-prop transport with per-domain contexts/hooks (`useAppShellCore`, `useAppShellActions`, route contexts), removing the system route-prop plumbing.
- **#185** — Extracted the push registration lifecycle out of the app shell into its own hook/module.
- **#181** — Unified the Nostr inbox event processing into a single pipeline shared by inbox sync and chat sync.

Closes #186
Closes #184
Closes #188
Closes #187
Closes #178
Closes #183
Closes #180
Closes #179
Closes #182
Closes #185
Closes #181

## Reviewer notes

All previously listed review findings have been fixed and folded into the per-issue commits above. No open findings remain, and none were rebutted.

Fixed since the last review pass:

- **#178**: authentication resolution now waits for native secret storage. `useCurrentNsec` exposes an `isResolved` flag (immediately true off-native, set once the platform bridge answers on native), and `AppShell` renders a blank page until it flips — so auth state is never decided without consulting native secret storage.
- **#182**: the duplicated persistence callbacks are gone. Both call sites (`useSaveCashuFromText`, `useNpubCashClaim`) now build their acceptor from the shared `createDefaultCashuTokenAcceptor` factory instead of repeating the persistence wiring.
- **#182**: restored the explanatory comment for the cashu-ts `batchRestore(300, 100)` budget on the funds-recovery path, so the numbers are no longer bare magic values.
- **#181**: the inbox pipeline restored the known-contact preference when resolving the message peer. Candidate pubkeys are now walked in order and the first one that resolves to a known conversation wins, with the unknown-thread path used only as a fallback — a message from a saved contact carrying multiple `p` tags can no longer be misfiled or misattributed.
- **#181**: kind-14/15 messages without a strict 64-hex inner rumor id are persisted again rather than dropped, and kind-7 reactions regained the outer-wrap-id fallback, matching what both previous per-path processors did.

Pre-existing test failures: none — the full unit suite passes on this branch.

## Post-review hardening (00dcb54)

A second deep review pass (two independent reviewers per area, old-vs-new behavior comparison) was run on top of the batch; all confirmed findings are fixed in the final commit:

- **Auth gate resilience**: the native nsec read behind the new auth gate is guarded with catch + 5s timeout falling back to the local snapshot, so a rejected/hung native bridge can no longer leave the app permanently blank.
- **Payment recovery scoping**: the unpublished-token recovery persist in `usePayContactWithCashuMessage` is scoped to the publish call itself; a post-publish error can no longer re-persist an already-sent token as pending.
- **Inbox pipeline parity**: restored pre-unification `created_at` coercion (non-positive/fractional → now), reactions resolved via in-flight references now join on the stable rumor id (reacting to an edited message renders instead of inserting an orphan-pruned row), and the accidental inner p-tag requirement for kind-7/5 was dropped (outer-wrap addressing is already enforced).
- **Dead code**: deleted the now fully-dead `topupMintClaim.ts` + test and the producers orphaned by #179 (`requestDeriveNostrKeys`, unconsumed exports).

Each behavioral fix has a regression test; `bun run check-code` passes and the full unit suite is green (510 tests).


jarvisonsmart:
## Local test report

Tested the PR at head `00dcb54`. No application regression was reproduced, and the worktree remained clean after testing.

### Automated checks

- `bun run build` — passed
- `bun run test` — passed: 510 web-app tests + 22 core tests
- `bun run check-code` — passed: typecheck, ESLint, and Prettier
- App-shell Playwright suite at the intended 390×844 mobile viewport:
  - Chromium: 5/5 passed
  - Firefox: 5/5 passed
- Real test-mint Cashu send in Chromium: 1/1 passed
  - funded 100 test sats
  - selected the eligible mint
  - split/persisted the 10-sat send and change
  - published the Cashu DM and payment notice

### Multi-browser / two-user testing

Ran two isolated local users in Chrome and Firefox against the local Vite app.

Verified:

- plain Nostr messages delivered in both directions
- active-chat subscription received new messages
- global inbox persisted a message while the recipient was on the wallet screen
- 10-sat and 1,000-sat Cashu messages sent from Chrome reached Firefox
- the recipient accepted and persisted both tokens; its wallet reached 1,008 sats after proof fees
- reply, edit, reaction toggle, copy, and archive-to-unknown behavior
- SLIP-39 restore and logged-out auth gating in both browser engines

### Settings/context coverage

The migrated system-settings contexts were exercised through the UI:

- all 12 relevant routes rendered without an error boundary: settings, advanced, master keys, auto-pay limit, push debug, mints, Nostr relays/new relay, Evolu servers/new server, current data, and history
- `Prefer cashu` survived disable → reload → re-enable
- the selected main mint could be changed and restored
- temporary Nostr and Evolu servers could be added and removed

### Findings

No product defect was found.

One test-infrastructure issue is worth fixing: the checked-in `playwright.config.ts` uses the desktop default viewport, while `appshell-parity.spec.ts` and `token-send.spec.ts` expect the mobile-only `Wallet` navigation button. Under the current config they time out before reaching the intended flow. Both suites pass when run with a 390×844 viewport. The viewport should be set in those specs or in a mobile Playwright project.

The live cross-mint melt was only partially verifiable. `testnut.cashu.space` uses a fake Lightning wallet and reported its melt successful without settling the real `cashu.cz` invoice. Linky retained the 18-sat source remainder and persisted the 987-sat target claim for background recovery, so the recovery boundary behaved correctly, but completion needs two real interoperable mints.

Remaining device/external coverage should focus on:

- native secret-storage rejection/hang fallback on Android/iOS
- real Web Push and native FCM registration lifecycle
- real-mint cross-mint melt plus crash/reload pending-claim recovery

PR #204, opened 2026-08-01
Durable notifications: shade delivery, ≥10s in-app banner, notification centre, and sync completeness

PeterXMR:
## The bug

> "notification showed just for 2 or 3 second and than disapeared, I wasnt fast enough to write whole content. Than I didnt saw any notification so I didnt saw any notifications afterwards."

Three separate defects were behind that report, and a fourth was found while fixing them:

1. **The toast was transient.** `useToasts` hardcodes a 2500 ms timeout (measured on-device at 2538 ms) and caps at two visible; a third arrival evicts the oldest **and clears its timer**, destroying the content with no affordance.
2. **Nothing durable was ever written.** Visibility, focus and open-chat checks ran *before* any persistence, so an event arriving in the wrong state left no trace at all — not a missed banner, an unrecoverable loss.
3. **The Capacitor WebView has no `Notification` API**, so on Android nothing reached the notification shade from the app path.
4. **Bootstrap sync could not fetch everything.** `querySync(..., { limit: 50 })` with no `since`, against outer timestamps NIP-59 randomises up to two days into the past. Measured on the real relays: the bounded query returned **50** where the unbounded one returned **172**, and only **1 of 5** test wraps fell inside the window.

## The fix

**Android durability.** A `@JavascriptInterface` bridge posting `NotificationCompat` with `MessagingStyle`, keyed on `(tag, id)` so re-posting updates one entry instead of stacking. An in-process cache is the sole append authority — `getActiveNotifications()` never appears on the append path, which is what makes bursts deterministic. `when` is always receipt time, never the event's: Android's `PeekOldWhenSuppressor` silently refuses heads-up for anything older than 24 h, and NIP-59 backdates.

Shade posts are coalesced (leading edge + trailing flush + settle). Android rate-limits to ~5 enqueues/sec per package and **sheds** the excess; five messages posting a child and a summary each is ~10 enqueues in 17 ms. Before coalescing, five rapid messages rendered five lines on only 3 of 10 runs.

**A durable record store.** Owner-scoped, debounced, merge-on-flush. Three timestamps rather than one boolean, mirroring Signal: `deliveredAt` always set, `alertedAt`, `readAt`, plus `dismissedAt` which **never** implies `readAt` — clearing the shade must not clear unread state. The write happens **before every alert gate**, so a message arriving while its own chat is open is recorded and merely not alerted.

Three dedup gates were conflating "this device has the MESSAGE" with "this device has a RECORD" — independent stores — and became fall-through flags that suppress only the Evolu write.

**A ≥10 s in-app banner.** No Android API controls heads-up duration (AOSP `heads_up_notification_decay` is 5000 ms, OEM-overridable; `setTimeoutAfter()` *cancels* the notification). Measured heads-up dwell here: **5.66–7.90 s**. So durability is Android's job and duration is the banner's. Dwell is anchored to paint via `useLayoutEffect` — armed at enqueue it measured 9646 ms, because the card reaches the DOM one commit later. Per-sender collapse bounds worst-case drain to `distinct_senders × 10 s`.

**A Notifications page.** `#settings/notifications`, reachable from a settings row carrying a live unread badge. Newest-first, per-item mark read **and** unread, "Mark all read" in the header which also clears Linky's shade entries — scoped to Linky's own group, never `cancelAll()`.

**Sync completeness.** The bootstrap `querySync` is replaced by a single `since`-bounded `pool.subscribe` whose EOSE both classifies delivery origin and advances a persisted per-owner watermark. Origin is then confirmed against the **inner rumor** timestamp, because EOSE alone classifies by when a byte arrived — a property of the relay, not the message — and relays replay newest-outer-first while nostr-tools fires `oneose` on a 4400 ms timeout.

## Verified on Pixel_6_API_36_Linky

Every phase carries its own device gate; the end-to-end pass is Section A of `VERIFY-01`.

| Property | Measured |
|---|---|
| Banner dwell | **10 163 ms** (was 2538 ms) |
| Burst of 5, one sender | 5 stacked lines, **10/10 runs** |
| 3 senders | 3 entries + group summary, no overwrite |
| Message while its chat is open | Record written, **3/3 consecutive runs** |
| Kill app → 10 messages → reopen | **10/10 present, send order, 0 duplicates** |
| Catch-up alerts | **0 banners / 0 shade posts**, both faithful runs |
| Old-timestamp event | Still heads-ups |
| Tap a notification | Opens that chat, cold and warm start |
| Mark all read | Badge to 0 and 6 shade entries to 0, one tap |

Zeros are non-vacuous: the shade sampler read 490 consecutive zeros, then its first non-zero tick lands on the exact second a live control published, in the same sampler instance.

## Tests

**1126 web-app** (was ~800), 22 core, 9 push, **43 Android JVM** via a new `android:test` (runs without `google-services.json`). Typecheck clean across six workspaces; `eslint` and `prettier --check` clean without any autofixer.

## Merge note

This branch merges `origin/main` (through `62f33d9`). #181's `nostrInboxPipeline.ts` turns four gates into hard `return ignored(...)`, including `known-message-wrap` and `ownsConversation` — which would reintroduce "no record while the chat is open". The pipeline is extended **additively and default-off** via `policy.reportsSuppressedWrites`; without that flag its behaviour is unchanged, so upstream's pipeline tests and `useChatNostrSyncEffect` are unaffected.

## Known limitations, measured not assumed

- **Relay ceiling.** The window is watermark-anchored and widens with time away. Wider than a relay's own `maxFilterLimit` (strfry default 500) it re-truncates on the same randomised axis. Measured **172/500** for the test identity — a 2.9× margin that tightens with use. Bounded backward `until`-pagination is designed and held in reserve.
- **Live-classification allowance is 30 s**, against a measured backlog gap of 38–62 s — narrowest headroom **8 s**. Closes the measured failure; does not bound the general case. A sender whose clock is >30 s slow is recorded but not alerted: a missed alert, never a missed message.
- **FCM paths are unverified this milestone.** A debug build has no `google-services.json`, so no Linky code runs while the process is dead; closed-process delivery and the placeholder-tag reconciliation are release-build-only. Recorded as a named gap, not implied by a debug pass.
- **Browser PWA convergence is deferred** by the repo owner's decision. The reported bug and every reproduction were on native Android.
- Czech copy wants a native speaker's review.


hynek-jina:
This is quite big without clear purpose.. Closing for now..
@PeterXMR , please rather start with issue, so we can agree on basic requirements first 🙏

PR #211, opened 2026-08-03
Fix bank payment offers stalling until app restart

kaladivo:
## Problem

When a recipient accepts a proxy (bank) payment offer, the author's app sometimes never sends `bank_details_sent`, leaving the acceptor waiting indefinitely. Restarting the author's app sometimes helped.

## Root causes

1. **The live inbox subscription silently dies.** `nostrPool.ts` created `new SimplePool()` with defaults — `enablePing` and `enableReconnect` are both off in nostr-tools. When a relay websocket drops (mobile background, wifi→LTE switch, relay idle timeout), all live subscriptions close permanently and the `accepted` event never arrives. While the app is open, the service worker also suppresses the push notification in favor of in-app handling, which depends on the same dead subscription — so the author gets no signal at all.
2. **Restart backfill could miss the accepted wrap.** The catch-up query used `limit: 50` (newest first by `created_at`). NIP-59 randomizes the outer wrap timestamp up to 2 days into the past, so a freshly published `accepted` wrap can rank below 50 other wraps and never be returned — the "restart doesn't always help" case.
3. **No retry after a failed send.** The auto-responder effect only re-runs on message-state changes. If the `bank_details_sent` publish failed once (offline relays, skipped lease lock), nothing scheduled a retry.

## Fix

- Enable `enablePing` + `enableReconnect` on the shared pool. Ping detects half-dead sockets (browser fallback uses a dummy REQ) and reconnect resumes open subscriptions.
- Backfill inbox wraps with a 3-day `since` window (matching the push service's catch-up window) and `limit: 500`, and re-run the backfill when the app returns to foreground or regains connectivity — the pool's reconnect resume filters by last emitted `created_at`, which NIP-59 timestamp randomization defeats, so missed wraps still need the re-query.
- Retry the offer auto-responder every 30 s while a non-expired `accepted` entry of the author's own offer is still waiting for bank details. The retry stops once details are sent, the offer ends, or the 5-minute phase TTL expires.

## Testing

- `bun run check-code` passes (typecheck, eslint, prettier)
- `bun run test` passes (113 files, 569 tests, incl. `useInboxNotificationsSync` suite)


kaladivo:
Addressed review findings (gpt-5.6-sol review, verified before applying):

1. **Dead-on-arrival subscription never revived** — the pool never retries relays that fail to connect at subscribe time, so an app started offline stayed deaf even after the online-event backfill. Each refresh now checks `listConnectionStatus()` and re-subscribes when any configured relay is disconnected.
2. **Transparent relay reconnects can skip wraps** — the pool resumes subscriptions with `since = lastEmitted + 1`, which NIP-59 timestamp randomization defeats, and no foreground/online event fires in that scenario. Added a 5-minute periodic refresh while the app is visible.
3. **Backfill cost on foreground flapping** — refreshes are now throttled to at most once per minute.

Skipped (judged not worth the complexity): keeping ping/reconnect alive for relays removed from the user's config mid-session — a rare, session-scoped idle connection with negligible cost; would need reference-counted per-relay cleanup on the shared pool.

PR #225, opened 2026-08-04
e2e test for proxy payment

bitcoinpolis:
Does not touch prod code, only add e2e test which runs the full happy path of proxy payment, all services locally (docker compose setup from @kaladivo )

You can run the test with:
```
docker compose -f docker-compose.dev.yml --profile e2e up -d --build --wait

cd apps/web-app
bunx playwright test --project=local-stack                      
```


docker-compose.dev.yml
```diff
@@ -58,6 +69,32 @@ services:
...
+        VITE_NPUB_CASH_DISABLED: "true"
+        VITE_PUSH_SERVER_URL: http://localhost:8787
+    ports:
+      - "5176:80"
+    healthcheck:
+      test: ["CMD-SHELL", "wget -qO- http://127.0.0.1/ >/dev/null || exit 1"]
+      interval: 3s
+      timeout: 2s
+      retries: 20
+      start_period: 3s
```
kaladivo, line 95: Would be nice to move this into separate script. Most of the time when developing we don't want to start the app container too.

docker-compose.dev.yml
```diff
@@ -7,6 +7,17 @@ services:
       - "7777:8080"
     volumes:
       - nostr-relay-data:/usr/src/app/db
+    # Without this, `up -d --wait` returns when the container is running, not when
+    # the port accepts — and a relay that is not connected at subscribe time is
+    # never retried by the app, which strands the E2E status/offer flow.
+    # `bash` explicitly: /dev/tcp is a bash builtin and the image's /bin/sh is dash.
+    # The image ships no curl/wget, so this is the available probe.
+    healthcheck:
```
kaladivo, line 16: Nice - forgot about this. Good call!

kaladivo (commented):
Mostly nitpicks

docker-compose.dev.yml
bitcoinpolis: Will let my Claude know ;-)

docker-compose.dev.yml
kaladivo: Oh, it's in e2e profile, my comment was wrong.

PR #235, opened 2026-08-08
Use full image name, solves 'did not resolve to an alias' and increas…

petrkr:
In some environments without aliased containers.conf it does not find those containers, so use full path.

As side-effect it increase security as long you can not use  container injection by relacing default 'docker.io' with some different registry, which may have infected containers.


kaladivo:
Lgtm good change

PR #239, opened 2026-08-08
Migrate from Bun to pnpm + Node (drop Bun completely)

kaladivo:
## What

Three commits that together remove Bun from the repo entirely and standardize the toolchain on pnpm + Node 24:

**1. Package management → pnpm** (`pnpm@11.20.0`)

- `bun.lock` → `pnpm-lock.yaml` + `pnpm-workspace.yaml`; root scripts use `pnpm --filter` / `pnpm -r --if-present run`
- CI: `oven-sh/setup-bun` → `pnpm/action-setup@v4` + `actions/setup-node@v5` (pnpm cache) in tests, e2e, and both Android workflows; `bunx playwright` → `pnpm exec playwright`
- `docker/web-app/Dockerfile` is pure node+pnpm
- `patch-android-java.sh` globs `node_modules/.pnpm/...`; Capacitor android/ios projects regenerated via `cap sync` (hence the generated-file churn incl. Capacitor 7.6.0 → 7.6.8 within the existing `^7.4.3` range)

**2. Push service runtime → Node** (was `Bun.serve` + `bun:sqlite`)

- Runs TypeScript sources directly via Node's type stripping — no build/bundle step; relative imports carry explicit `.ts` extensions for ESM resolution
- `Bun.serve` → `node:http` with a small adapter feeding the existing fetch-style handler (handlers themselves unchanged; client IP comes from the socket instead of `Bun.Server.requestIP`)
- `bun:sqlite` → `node:sqlite`. Statements reading rowids use `setReadBigInts(true)` because node:sqlite throws on integers above `Number.MAX_SAFE_INTEGER` where `readSafeInteger` must reject the row instead (the existing overflow hardening tests cover this); a BEGIN/COMMIT/ROLLBACK helper replaces `db.transaction`
- `bun:test` → vitest, so the root `pnpm run test` now covers the push workspace too
- `web-push` switches to a default import (Node's CJS interop exposes no named exports for it — caught by a live smoke test, not typecheck)
- Docker image runs the sources via `pnpm deploy --prod --legacy` (pruned to the 3 prod deps); local `dev`/`start` load `.env.development` via `node --env-file*`, matching Bun's implicit env loading

**3. Node 24 everywhere, pinned by `.node-version`**

- `.node-version` (`24`) drives version managers and CI (`setup-node` reads it via `node-version-file`); `engines: { node: ">=24" }` in root + push manifests warns on mismatched local node
- All Docker images on `node:24` (push, web-app e2e build, evolu-relay dev image)
- On 24 (active LTS), `node:sqlite` and type stripping are stable — the push service starts with zero experimental warnings
- pnpm `engineStrict` was tried and dropped: it also enforces *dependencies'* engines, and `slip39-ts` ships stale metadata (`^20.9.0 || ^22.2.0`) that rejects 24 despite being a browser-targeted lib

## Kept behavior-neutral

Fresh resolution pulled two packages whose new versions change output, so they are pinned to what `bun.lock` had — upgrading them is deliberately left for separate PRs:

- `prettier` pinned to `3.8.1` (3.9.x reformats union types across ~18 source files)
- `eslint-plugin-react-hooks` pinned to `7.0.1` (7.1.x adds `set-state-in-effect`, which flags `apps/site/src/cashu/CashuPage.tsx:1832`)

## Verified locally (pnpm 11.20.0, node 24.18/24.19)

- `pnpm run check-code` and `pnpm run test` pass under Node 24 (608 tests incl. the 5 push tests now on vitest)
- `local-stack` e2e suite passes against the web-app image built from the new node:24 Dockerfile
- Push service smoke-tested live on Node 24: boots with zero warnings, connects to relays; `/health`, `/`, `/vapid-public-key`, `/auth/challenge` (SQLite write path), invalid-JSON handling, CORS preflight, and 404 all behave as before
- Push Docker image (node 24.19) builds, boots, and fails fast on missing required env; evolu-relay dev image boots and accepts websocket connections on 24

## After merge

- Anyone with a checkout: switch to Node 24 (version managers pick up `.node-version`), `pnpm install`, plus `pnpm exec playwright install chromium` once (Playwright patch bump = new browser build); Bun no longer needed
- If the Vercel projects have a custom `bun install` install command in the dashboard, switch it to `pnpm install` (auto-detection picks pnpm up from the lockfile otherwise); Vercel project Node version should be set to 24 if it is pinned lower in the dashboard


kaladivo:
There is a new version of bun that addresses a lot of shortcommings. 

Closing this

Issue #247, opened 2026-08-14
Audit the private image / Blossom path


kaladivo:
The chat image pipeline (\`apps/web-app/src/app/lib/privateImageMessage.ts\`) has grown organically and deserves a focused audit before we build more on top of it.

Things to look at:

- **Single hardcoded upload server** (\`blossom.primal.net\`) with the app's \`/api/blossom-upload\` proxy as the only fallback. No mirror/retry across servers; one outage breaks image sends. Should the server list be configurable or multi-target?
- **Ciphertext stored as base64url text** (uploaded as \`text/plain\`), inflating every blob by ~33%. Presumably for proxy/server compatibility — verify whether raw binary upload works against current Blossom servers and whether the \`storageEncoding: "raw"\` path is actually exercised anywhere.
- **kind 24242 auth event**: check tag set and expiration handling against the current BUD-02 spec, and that the proxy fallback still validates auth server-side.
- **Crypto sanity**: AES-GCM key/nonce generation, key/nonce travel in rumor tags (kind 15) — confirm this matches the NIP-17 file-message shape other clients expect, or document where we deviate.
- **Verification**: size + ciphertext hash + plaintext hash are all checked on receive — good; confirm the upload path can't accept a server-rewritten URL pointing elsewhere.
- **Cleanup/orphans**: uploaded blobs are never deleted; failed sends can leave orphaned blobs on the server. Is that acceptable?
- **Service worker interplay**: the SW has a CacheFirst route for image destinations — check it doesn't cache encrypted blobs in a way that breaks the size/hash checks or leaks ciphertext lifetime expectations.

Related: the send-side rumor construction is moving to the linkstr chat vertical (kind 15 wire format already mirrored there); this audit is about the HTTP/media half that stays app-side.

PR #250, opened 2026-08-14
Fix Nostr avatars reloading, and batch profile relay queries

PeterXMR:
## Problem

Reported: *"images often reload even though I already had them. Sometimes the image doesn't load. Overall, check to see if I'm calling the relays incorrectly."*

All three turned out to be real, and present on current `main`. I checked merged PRs first — #121 *avatar refresh* (2026-05-27) introduced this caching layer and nothing since has touched the churn.

## Root cause: an object-URL cache that was never read

`avatarObjectUrlByNpub` (was `avatarObjectUrlsByNpubRef`) tracked the live `blob:` URL per npub, but was only ever **written** — nothing read it to decide whether a new URL was needed. So every prefetch pass:

1. called `URL.createObjectURL(blob)` again for an avatar already on screen,
2. `rememberBlobAvatarUrl` revoked the **previous** URL — the one still in the ``,
3. React saw a changed `src` → browser drops the decoded image and reloads it.

When the revoke landed while that reload was still in flight, the load failed outright — that is the "sometimes the image doesn't load" half.

The loop ran much more often than it appears: `prefetchedMessageNpubs` is memoized on `chatMessages`, so **every incoming chat message** re-ran the entire avatar pass, for every unknown-chat participant.

Because `canFetchAvatarAsBlob` only permits blob caching for same-origin and `api.dicebear.com`, this hit precisely the DiceBear avatars Linky generates for new accounts — i.e. most Linky-to-Linky contacts.

## Fixes

**1. One object URL per npub.** Lifetime now has a single owner in `nostrProfile.ts`. `loadCachedProfileAvatarObjectUrl` returns the URL it already handed out instead of minting a second one; only `cacheProfileAvatarFromUrl` (genuinely new bytes) and `deleteCachedProfileAvatar` replace or release it. `rememberBlobAvatarUrl` no longer keeps a competing registry.

**2. An empty relay response no longer blanks a good avatar.** Unreachable relays and a genuinely absent profile are indistinguishable from the client, but the code cached both as "this profile has no picture" — so one slow round blanked avatars already on screen, then refetched 2 minutes later. `recordProfileMetadataLookup` / `recordProfilePictureLookup` record a miss only when nothing is cached yet.

**3. Relay queries are batched.** This is the "calling the relays incorrectly" part. Prefetch issued **one `querySync` per contact, awaited sequentially**, each with an 8s `maxWait` — 50 contacts meant up to 50 serialised round trips, which is why avatars trickled in one at a time on a cold start.

kind:0 is a replaceable event, so a single filter can carry many authors and the relay returns the newest per author. `primeProfileMetadataCache` warms the cache in one multi-author query (chunked at 50, the widely-safe filter cap), with in-flight dedupe so concurrent effects share a query. The existing per-npub path stays as the fallback and for the 12h staleness refresh.

## Tests

`src/nostrAvatarObjectUrl.test.ts` (6) — pins URL identity across repeated loads, per-npub isolation, revoke-only-on-replace, single and bulk release. Written first; it failed with `expected 'blob:mock/3' to be 'blob:mock/2'`, reproducing the churn.

`src/nostrProfileBatch.test.ts` (11) — single multi-author query, cached npubs skipped, no query when fully cached, concurrent callers share one query, empty/rejected responses record no misses, and the recorders never downgrade a known-good value.

## Verification

- `bun run check-code` — clean (typecheck, eslint, prettier)
- `bun run test` — 21 tests across the three profile suites pass

`src/app/hooks/mint/useNpubCashMintSelection.test.tsx` has 3 failures; I confirmed by stashing that these fail identically on clean `main` and are unrelated to this change.

Not covered here: chat images (Blossom) are a separate pipeline — see #247.

### DIFF
--- a/apps/web-app/src/app/hooks/composition/useContactsMessagingComposition.ts
+++ b/apps/web-app/src/app/hooks/composition/useContactsMessagingComposition.ts
@@ -24,6 +24,11 @@ import {
   loadCachedProfileMetadata,
   loadCachedProfilePicture,
   NOSTR_RELAYS,
+  primeProfileMetadataCache,
+  recordProfileMetadataLookup,
+  recordProfilePictureLookup,
+  releaseAllAvatarObjectUrls,
+  releaseAvatarObjectUrl,
   saveCachedProfileMetadata,
   saveCachedProfilePicture,
   type NostrProfileMetadata,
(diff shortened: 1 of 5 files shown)


kaladivo:
This won't be merged due to a big nostr rewrite currently in progress. 

Will implement your intention as a part of the rewrite :). 

see https://github.com/hynek-jina/linky/pull/241

PeterXMR:
Thx for feedback, seems like I missed that one :)

kaladivo:
Thanks for the deep dive @PeterXMR — the diagnosis was spot-on, and even though this PR targets the old pipeline (which the linkstr migration on `linkstr-reactions` has since replaced), all three fixes have been carried over into the new one:

- **Object-URL churn (fix 1):** `profileCache` is now the single owner of avatar object URLs — one live URL per npub, re-loads return the URL already handed out, revocation only when the bytes are replaced, the avatar is deleted, or the identity changes (6e56270). Your `nostrAvatarObjectUrl` tests were adapted almost verbatim to pin those semantics.
- **Empty responses blanking avatars (fix 2):** the new pipeline is watch-fed and event-driven; caches are newest-wins by event `created_at` and there is no "record a miss" path left, so this one can no longer occur structurally.
- **Relay batching (fix 3):** one-shot profile lookups go through a chunked multi-author query (`Profiles.fetchProfiles`, 50-author filter cap — your number), the live `ProfileWatch` subscription chunks its authors filter the same way, and contact-suggestion discovery got the same treatment (ae87371, 1e44c65, f8bd3de).

So this stays closed as superseded rather than rejected — it saved us from re-discovering all three the hard way.

Issue #222, opened 2026-08-04
Dev server shows the PWA update banner on every service-worker change


kaladivo:
## Behavior

When running the app via `bun run dev` / `bun run dev:prod`, the "update Linky" banner appears in the app. Users running a local dev server get prompted to update as if a production release had shipped.

## Why it happens

Dev mode intentionally keeps the PWA service worker registered (`devOptions.enabled: true`, used for push testing). The plugin is configured with `registerType: "prompt"`, so whenever the built dev SW changes by even one byte (any edit to `src/sw.ts` or its imports), the browser installs the new worker into "waiting" and the app shows the same update banner that is meant for production releases.

Besides being confusing, the waiting worker keeps serving stale behavior until the prompt is accepted — e.g. after the dev-inspector change, the old SW's navigation fallback kept hijacking `/inspector.html` until the update was applied (a one-time reset via `#advanced/push-debug` also works).

## Expected

In dev, SW updates should activate silently (no banner). A likely shape: in dev builds, call `self.skipWaiting()` on install + `clients.claim()` on activate (or auto-post `SKIP_WAITING` from the client when `import.meta.env.DEV`), keeping the prompt flow production-only. Needs a quick check that auto-activation doesn't disrupt the push-debug workflow that dev SW persistence exists for.

## Repro

1. `bun run dev`, open the app, let the SW register
2. Touch `src/sw.ts` (any change), reload the app
3. Update banner appears

PR #251, opened 2026-08-14
Gate browser deploys on a version bump so the update banner only announces real releases

PeterXMR:
## The bug

"Sometimes an update is offered even though I'm already on the latest version" — `registerType: "prompt"` fires `onNeedRefresh` whenever the service worker script differs by a byte, and the SW embeds the precache manifest, so *any* redeploy rewrites it. Since we auto-release every merge to main, a deploy without a version bump announced "a new version of Linky is available" while the version in Settings stayed identical.

## The fix (reworked per review feedback)

The first commit fixed this client-side by asking the pending worker for its version. Review feedback rightly pointed out that if we never build without bumping the version, no client check is needed — so the second commit reverts all of that and standardizes the release process instead:

**A browser release is a version bump.** A Vercel Ignored Build Step (`apps/web-app/vercel.json` `ignoreCommand` → `apps/web-app/scripts/vercel-ignore-build.sh`) compares the root `package.json` version at `HEAD` against `VERCEL_GIT_PREVIOUS_SHA` (the last successful deployment) and skips the build when unchanged. With no build without a bump, the SW manifest can only change alongside a real release, so the banner can only ever be truthful — by construction, with zero client code.

Every failure path (first deploy, shallow clone missing the previous commit, unreadable version) exits toward *building*, so the gate can never silently withhold a release.

## Trade-off to be aware of

Merges to main without a bump now deploy **nothing** to browsers until the next CalVer `MICRO` bump — shipping a hotfix means bumping the version. That is the standardization proposed in review, stated here explicitly.

⚠️ If the Vercel project's Root Directory is not `apps/web-app`, the `vercel.json` is not picked up and the `ignoreCommand` must be set in the project settings instead.

## Verification

- Net diff vs main is only the gate: the client-side files are byte-identical to `origin/main` again (`git diff origin/main` shows +60 lines across 3 files)
- `bun run check-code` passes; the 7 pre-existing `pwaUpdate` tests pass
- The ignore script was exercised on all four branches against real repo history: same version → skip (exit 0); bump `0.0.0 → 26.9.0` → build; empty `VERCEL_GIT_PREVIOUS_SHA` → build; unknown SHA → build

## Note on #222

Dev rebuilds are untouched by this (the gate only runs on Vercel), so the dev-server banner half of #222 stays open.


PeterXMR:
> Your trial has ended. [Reactivate Greptile](https://app.greptile.com/-/pull-requests) to resume code reviews.

Not so useful comment, not paying a penny for this tool personally :)

kaladivo:
I am not sure this is really an issue here. 

We are currently doing a bit "hasty" releases but that is by design - we automatically release from main. So in this case release != version bump (for browsers only). 

This needs to be thought trough more honestly. If we don't want to release main to browsers, we should not run build at all without bumping the version. If we don't run build at all without bumping the version we don't need to check the version. 

From my perspective this just adds unnecessary code - but we should definitely standardize release process that is a very good insight. Feel free to draft that out :) if you have some idea about it :)

kaladivo:
> > Your trial has ended. [Reactivate Greptile](https://app.greptile.com/-/pull-requests) to resume code reviews.
> 
> Not so useful comment, not paying a penny for this tool personally :)

We need to active it, they have a opensource program. It's quite good honestly

PeterXMR:
> I am not sure this is really an issue here.
> 
> We are currently doing a bit "hasty" releases but that is by design - we automatically release from main. So in this case release != version bump (for browsers only).
> 
> This needs to be thought trough more honestly. If we don't want to release main to browsers, we should not run build at all without bumping the version. If we don't run build at all without bumping the version we don't need to check the version.
> 
> From my perspective this just adds unnecessary code - but we should definitely standardize release process that is a very good insight. Feel free to draft that out :) if you have some idea about it :)

@hynek-jina mentioned an issue in PM
"Sometimes an update is offered even though I'm already on the latest version"
 and this PR fix it.
 
 Also have in mind what you just mensioned now as I rewrirte it

PR #272, opened 2026-08-15
Restore contact names clobbered by the profile-follow sync

kaladivo:
## What happened

PR #241 switched contact names from "fill once when empty" to "follow the Nostr profile unless `contact.nameSetByUser`". The flag shipped without a backfill, so every pre-existing contact had it unset and the first profile event after the 2026-08-15 deploy overwrote every manually chosen contact name with the profile's kind-0 name.

## The repair

A **one-time migration**, fully contained in `apps/web-app/src/app/migrations/` (plus one marked mount line in `useContactsMessagingComposition.ts`) so it can be deleted in one sweep later:

- `evolu_history` still holds every overwritten value, so nothing was actually lost. The migration reads the `contact.name` / `contact.nameSetByUser` history rows straight from it.
- **Detecting clobbers:** Evolu gives all column writes of one mutation a single shared HLC timestamp, and the contact editor always writes `name` and `nameSetByUser` together. So a `name` write at/after the PR #241 merge time with no same-timestamp `nameSetByUser` write can only be a profile-sync overwrite.
- **Restoring:** for each contact without `nameSetByUser` whose latest history is a sync clobber, the newest non-empty pre-deploy name is written back together with `nameSetByUser = true`, so the sync leaves it alone from now on.
- **Skipped on purpose:** contacts the user has renamed or explicitly reset to the profile since the deploy (post-deploy editor writes), names that were merely filled into an empty field (that was desired), and rows where nothing changed.
- The restore mutation syncs to the account's other devices through Evolu; restored rows carry the flag, so concurrent backfills on several devices converge. A `linky.contact_name_backfill_v1` localStorage marker keeps re-runs off the query path.

One behavioral note: restored names are frozen (flag set), including ones that originally came from an old auto-fill — exactly the pre-#241 behavior. The editor's "reset name from profile" clears the flag for anyone who wants follow-mode back.

## Removal

Delete `src/app/migrations/` and its mount line once the user base has run it; every file is headed with a removal marker and `docs/architecture.md` records the plan.

## Testing

- 10 unit tests on the pure restore planner (clobber detection, HLC counter ordering, editor/reset skips, multi-contact)
- `bun run check-code` and the full `bun run test` suite (594 tests) pass


kaladivo:
Not working, closing

PR #275, opened 2026-08-16
Migrate push service to Linkstr

### LINKED ISSUE #253: Migrate the push service (apps/push) from direct nostr-tools usage to linkstr
The web app and service worker now run all nostr protocol through `@linky/linkstr` (#241), but `apps/push` still consumes `nostr-tools` directly:

- `src/relayWatcher.ts` — its own `SimplePool` with a reconnecting kind-1059 subscription (`["linky","push"]`-tagged wraps), hand-rolled `verifyEvent` checks, seen-event dedupe cache, and relay-hint extraction
- `src/ownership.ts` — verifies kind-27235 ownership proofs with `verifyEvent` (the client side of these proofs is now produced by linkstr's `makePushOwnershipProof`)
- `src/guards.ts`, `src/types.ts` — nostr-tools `Event` type

This duplicates logic linkstr already owns: relay subscription lifecycle/resubscribe (`WrapInbox` / `NostrTransport`), event authentication, and the push-marker wire convention (`LINKY_PUSH_MARKER_TAG`/`VALUE`).

Migration notes:

- The push service never decrypts wraps and has no user identity, so it doesn't fit `LinkstrIdentity`-based composition as-is — it needs the transport/subscription layer and codec-level validation without an identity, or a small identity-free composition entry
- Proof verification should share the kind-27235 wire shape with `httpAuth/codec.ts` so client and server can't drift
- `docker/evolu-relay` protocol-compat gotcha doesn't apply here, but the push service Docker image build must keep working with the workspace dependency

kaladivo:
## What changed

- add an identity-free `PushInbox` to `@linky/linkstr` with per-relay resubscription, rolling lookback filters, outer-wrap signature validation, delivery phase tracking, relay-hint extraction, and authenticated cross-relay dedupe
- move kind-27235 push ownership-proof verification into Linkstr's shared HTTP-auth codec
- migrate `apps/push` off direct `nostr-tools` usage while retaining SQLite dedupe across process restarts
- update the push Docker build to include the Linkstr workspace source
- document the new composition and add Linkstr/push integration coverage

## Why

The web app and service worker already use Linkstr as their sole Nostr protocol surface, but the push service still duplicated transport lifecycle, wire validation, marker conventions, and proof verification. This gives the identity-free server a purpose-built Linkstr entry point without requiring decryption keys.

## Impact

Push delivery remains generic and non-decrypting. Historical events are suppressed using Linkstr's per-relay backfill/live phase, while live push-marked wraps continue to fan out to web and Android subscriptions.

## Validation

- `bun run check-code`
- `bun run test` (920 tests passed)
- `bun run --filter @linky/push build`
- `docker build -f apps/push/Dockerfile -t linky-push-linkstr-test .`

Closes #253


packages/linkstr/src/push/PushInbox.ts
```diff
@@ -0,0 +1,132 @@
...
+            Effect.gen(function* () {
+              const nowSeconds = Math.floor(
+                (yield* Clock.currentTimeMillis) / 1000,
+              );
+              const lookbackSeconds = Math.floor(
+                Duration.toMillis(options.lookback) / 1000,
+              );
+              const filter: Filter = {
+                kinds: [GIFT_WRAP_KIND],
+                [`#${LINKY_PUSH_MARKER_TAG}`]: [LINKY_PUSH_MARKER_VALUE],
```
kaladivo: **Blocker: this filter is rejected by the production relays.** `#linky` is a multi-letter tag key; NIP-01 only defines `#<single-letter>` filters. relay.damus.io, nos.lol, and relay.primal.net all run strfry (verified via NIP-11), and strfry's `src/filters.h` (checked at the deployed tags 1.0.3 and 1.1.0) unconditionally throws `unindexed tag filter` for any `#key` longer than one letter and answers `CLOSED "bad req"`. The transport surfaces the close, `resubscribeForever` retries forever with backoff, and the service delivers **zero notifications** while `/health` stays ok.

It looks fine locally because nostr-rs-relay 0.10.0 (`subscription.rs:156`) silently *ignores* the key — dev degrades to the full kind-1059 firehose instead — and `PushInbox.test.ts:124-129` asserts this exact filter against a fake transport, locking the bug in.

Fix: drop the `#linky` key from the wire filter. `decodePushWrap` already re-checks the marker client-side, so removing it restores exactly the old (working) semantics.

packages/linkstr/src/push/PushInbox.ts
```diff
@@ -0,0 +1,132 @@
...
+                since: Math.max(nowSeconds - lookbackSeconds, 0),
+              };
+              let eoseSeen = false;
+              yield* transport.subscribe(
+                relay,
+                filter,
+                (raw) => {
+                  Queue.unsafeOffer(rawWraps, {
+                    delivery: eoseSeen ? "live" : "backfill",
```
kaladivo: **Reconnect backfill window permanently drops genuinely new wraps.** `eoseSeen` is per subscription *attempt* (reset by every resubscribe), `since` is recomputed as `now − lookback` each attempt, dedupe (`seenWrapIds`) is global first-arrival-wins, and the consumer records-then-drops backfill wraps. So: a relay socket drops (`enableReconnect: false` makes every drop a fresh attempt), a wrap published around the reconnect arrives pre-EOSE on that relay first → labeled `backfill` → written to the durable `seen_events` ledger and dropped; the live copy from another relay is deduped. The notification is lost permanently (ledger retention is 3d6h). `PushInbox.test.ts:131-138` currently pins this as *intended* behavior.

For fairness: the suppress-and-record ordering is byte-for-byte pre-existing on main, and the old code was worse in aggregate (global phase flag + forced 3-day replay 144×/day meant cross-relay rescue was impossible). But old `enableReconnect: true` resumed from `lastEmitted+1` with **no** backfill window per reconnect, so the per-disconnect window is a genuine regression.

Fix direction: a per-relay cursor for `since`, or suppress/record only on the delivery path so a live re-arrival can promote an earlier backfill sighting.

packages/linkstr/src/push/headless.ts
```diff
@@ -0,0 +1,56 @@
...
+              readRelays: config.readRelays,
+              writeRelays: [],
+            }),
+            config.transport ?? NostrTransportSimplePool,
+          ),
+        ),
+      ),
+    ),
+  );
+  const fiber = Effect.runFork(program);
```
kaladivo, line 69: **Every failure mode of the watcher is invisible.** Three layers compound: decode failures map to `Option.none()` with no log (the typed `PushWrapFailure` is discarded), `resubscribeForever` `Effect.ignore`s every relay error and close reason, and `Effect.runFork` here has no failure observer — Effect *does* log unhandled fiber failures, but at Debug level, below the default Info threshold (empirically verified on this repo's effect 3.19.19 under Bun: a failing `runFork` prints nothing). No `Logger` configuration exists anywhere in `apps/push` or linkstr.

Combined with the filter bug, the production symptom is: one `[push] opening linkstr relay watcher` line, then silence forever, with `/health` green. The old code logged malformed wraps with a reason, EOSE catch-up, and subscription closes.

Fix: log the `PushWrapFailure` on the decode `Left`, surface close reasons from the resubscribe loop, and provide a `Logger.minimumLogLevel` / unhandled-error log-level layer here.

apps/push/src/relayWatcher.ts
```diff
@@ -17,268 +16,40 @@ interface RelayWatcherOptions {
...
-    const since = Math.floor(Date.now() / 1000) - CATCH_UP_LOOKBACK_SECONDS;
-    this.liveDeliveryEnabled = false;
+    if (this.subscription !== null) return;
     console.info(
-      `[push] opening relay watcher subscription reason=${reason} since=${since}`,
+      `[push] opening linkstr relay watcher with ${CATCH_UP_LOOKBACK_SECONDS}s lookback`,
     );
-    this.subscription = this.pool.subscribeMany(
-      this.relayUrls,
+    this.subscription = watchPushInbox(
```
kaladivo, line 49: **A stalled REQ on a live socket never heals now.** The old 10-minute forced subscription restart is gone, and `resubscribeForever` only re-fires after an attempt *completes* (i.e. after `onclose`). A relay that keeps the websocket alive but loses subscription state (relay restart behind an LB, proxy half-close) reads as healthy forever: the keepalive ping opens a *new* REQ id each time (in Bun, `ws.once` is undefined so nostr-tools always takes the patched `limit: 1` dummy-REQ path), so it detects dead sockets but not a dropped long-lived subscription. Old code self-healed within 10 minutes; new code stays deaf on that relay until process restart.

Worth restoring a periodic refresh, or tying liveness detection to the actual subscription.

packages/linkstr/src/push/codec.ts
```diff
@@ -0,0 +1,85 @@
...
+ * authenticates its id, tags and ciphertext for safe routing and dedupe.
+ */
+export const decodePushWrap = (
+  input: unknown,
+): Either.Either<PushWrap, PushWrapFailure> =>
+  Either.gen(function* () {
+    const wrap = yield* decodeWrap(input).pipe(
+      Either.mapLeft((): PushWrapFailure => "malformed-wrap"),
+    );
+    if (!verifyEvent(wrap)) {
```
kaladivo, line 64: **Each wrap is now schnorr-verified ~6× instead of ~1×.** Three compounding causes: (1) the transport already verified the raw event and nostr-tools caches that under `verifiedSymbol` *on the event object* — but this verifies a freshly constructed Schema instance, so the cache is lost and this is a second real verification; (2) `NostrTransport.subscribeToRelay` passes no `alreadyHaveEvent`, so with 3 relays each unique wrap is parsed+verified per relay (the old `subscribeMany` short-circuited duplicates before JSON.parse); (3) in `PushInbox.processRaw`, dedupe runs *after* `decodePushWrap`, so even in-memory duplicates pay full curve decode + schnorr first. Also the marker check below sits after `verifyEvent`, so unrelated 1059s pay the verify too.

Today the PR is still a net perf win because the relay-side filter shrinks the input — but that evaporates once the `#linky` filter is removed (see the PushInbox comment), making this the dominant cost on the backfill path.

Cheap fixes, no semantic change: move the marker check above `verifyEvent`; check the untrusted `raw.id` against `seenWrapIds` before decoding (a forged colliding id can only cause a drop, no security loss); or wire `alreadyHaveEvent` through the transport.

packages/linkstr/src/push/PushInbox.ts
```diff
@@ -0,0 +1,132 @@
...
+interface RawArrival {
+  readonly delivery: InboxDelivery;
+  readonly raw: unknown;
+}
+
+const GIFT_WRAP_KIND = 1059;
+const DEFAULT_RESUBSCRIBE_DELAY = Duration.seconds(5);
+const SEEN_WRAPS_CAPACITY = 4096;
+
+const makeSeenWrapIds = () => {
```
kaladivo: ~40 lines here are char-for-char copies of `inbox/WrapInbox.ts`: `makeSeenWrapIds` + `SEEN_WRAPS_CAPACITY`, `RawArrival`, `GIFT_WRAP_KIND`, `DEFAULT_RESUBSCRIBE_DELAY`, the acquire-release queue, and the per-relay `eoseSeen` closure. The copy also dropped WrapInbox's comment explaining the correctness invariant (only *authenticated* wraps may be marked seen — PushInbox relies on the same invariant by only adding ids on the `onRight` branch).

`docs/architecture.md` describes this as one shared behavior, but it's now two independently-editable implementations — any fix to phase semantics, dedupe, or teardown has to land twice. Suggest extracting the shared internals (a `seenWrapIds` module and a per-relay phased subscription feed next to `internal/resubscribe.ts`) and letting each inbox supply only its filter and `processRaw`.

apps/push/src/relayWatcher.ts
```diff
@@ -17,268 +16,40 @@ interface RelayWatcherOptions {
...
-
-    this.seenEventIds.markSeen(eventId, nowMs);
-    return true;
   }
 
   pruneSeen(nowMs: number): void {
-    this.seenEventIds.pruneExpired(nowMs);
     this.storage.pruneSeenEvents(
       nowMs,
       Math.max(
```
kaladivo: `eventDedupeTtlMs` is now inert: its only remaining use is this `Math.max` against a 78 h constant floor (3-day lookback + 6 h margin), and the default is 24 h — so `PUSH_EVENT_DEDUPE_TTL_MS` (still documented in `apps/push/.env.example`) does nothing unless set above 78 h, and can only ever *raise* retention. An operator lowering it to shrink the dedupe window gets no effect.

Suggest deleting the option and its plumbing (config field, `RelayWatcherOptions`, `index.ts` wiring, `.env.example` line) and pruning with the constants directly — or renaming it to what it now is, a retention-floor override.

kaladivo (commented):
Reviewed the migration end to end (all findings below were adversarially verified against the deployed relay software, nostr-tools internals, and the pre-PR behavior on main).

**TL;DR: as-is, this breaks push delivery entirely in production while passing every test.** The `#linky` wire filter is rejected by strfry, which is what relay.damus.io, nos.lol, and relay.primal.net all run — and the failure is invisible because the watcher has no failure logging. Details inline. The three comments I'd treat as blockers: the wire filter, the missing failure logging, and the reconnect backfill-suppression race.

Also credit where due — verification showed the migration is a real improvement in several ways: the old watcher forced a full 3-day backfill window 144×/day with a global phase flag (so cross-relay rescue was structurally impossible), and the new per-relay design fixes that. A few scary-looking candidates were checked and *refuted*: a bad `PUSH_DEFAULT_RELAYS` entry cannot crash boot (config normalization provably satisfies the `RelayUrl` brand — fuzzed 400k inputs), and the ownership-proof path got cheaper per request, not more DoS-able.

**Why tests didn't catch the blocker, and how to fix that:** `PushInbox.test.ts` asserts the filter *shape* against a permissive in-memory fake, and the dev relay (nostr-rs-relay) silently ignores unsupported filter keys — so every layer validated the implementation, none validated the contract. Suggested layers: (1) make the fake transport strict — reject any non-single-letter `#` tag filter with CLOSED, like strfry does, so the unit test fails immediately; (2) add a strfry container to the e2e docker profile and assert actual wrap delivery through it, since that's what production runs; (3) surface subscription CLOSED reasons in logs (needed anyway, see the observability comment) so tests can assert their absence.

apps/push/src/relayWatcher.ts
```diff
@@ -288,112 +59,76 @@ export class RelayWatcher {
...
-      console.debug(
-        `[push] skipped malformed gift wrap id=${event.id} pubkey=${event.pubkey} recipients=${describeEventRecipients(event)} reason=${validation.reason ?? "unknown"}`,
-      );
-      return;
-    }
-
-    if (!this.recordSeen(event.id, nowMs)) {
-      console.debug(`[push] skipped duplicate event id=${event.id}`);
+    const nowMs = Date.now();
+    if (!this.storage.recordSeenEvent(wrap.wrapId, nowMs)) {
```
kaladivo, line 108: Two things on this line. First, ordering: `recordSeenEvent` runs before the `delivery === "backfill"` drop below, so a suppressed wrap is durably recorded and can never be delivered later, even across restarts — this is what makes the backfill race in `PushInbox` permanent. (The ordering is pre-existing on main, but it's newly load-bearing now that reconnects create backfill windows.)

Second, coverage: `relayWatcher.test.ts` was deleted and nothing replaced it — there's no test for this record-before-drop ordering, none for `SEEN_WRAPS_CAPACITY` FIFO eviction in `PushInbox`, and none for `pruneSeen`. The deleted suite pinned exactly this class of bug (eviction/TTL regressions); a refactor that reorders the ledger write or breaks eviction currently ships green.

docs/architecture.md
```diff
@@ -35,11 +35,12 @@ Architectural decisions and behavioral constraints for Linky. Keep this file up
...
+- `PushInbox` is the identity-free kind-1059 sibling used by `apps/push`: `watchPushInbox` composes it over `NostrTransport` without a `LinkstrIdentity`, one reconnecting subscription per relay filters `#linky=push` with a fresh rolling lookback, and the service emits signature-verified outer-wrap facts with recipient/relay-hint routing plus per-relay `backfill`/`live` delivery. Authenticated wrap ids are FIFO-deduped across relays in Linkstr; the push service retains only its SQLite seen-id ledger for restart-safe dedupe. It never decrypts wraps or imports `nostr-tools`
```
kaladivo: Two doc nits: "It never decrypts wraps or imports `nostr-tools`" — with `PushInbox` as the bullet's leading subject this reads as a claim about linkstr's push module, which *does* import nostr-tools (`push/codec.ts` imports `verifyEvent`; it's true only of `apps/push`). As written, someone "restoring" the invariant would delete the only signature check on the push path. Worth naming `apps/push` explicitly.

Also, the push-service paragraph further down (~line 182) still describes the deleted in-memory seen-id cache and the removed 10-minute periodic subscription refresh — should be updated in this PR.

apps/push/src/ownership.ts
```diff
@@ -12,46 +13,8 @@ interface OwnershipVerifierOptions {
...
-  return value;
-}
-
-function expectedProofContent(action: ProofAction): string {
-  return action === "subscribe"
-    ? "linky-push-subscribe"
-    : "linky-push-unsubscribe";
-}
+const failureStatus = (failure: PushOwnershipProofFailure): number =>
+  failure === "invalid-signature" || failure === "invalid-pubkey" ? 401 : 400;
```
kaladivo: Heads-up, not a blocker: this mapping shuffles the error contract. Non-hex `sig` went 400→401, wrong `action` tag 401→400, duplicated `pubkey` tag 400→401, and structurally-bad events collapsed from field-specific `invalid_request` messages to a generic `invalid_proof: malformed-event` — with internal failure tags now leaking into user-visible strings (the web client surfaces the raw body). Verified that no client branches on these codes today and every previously-rejected shape is still rejected, so the cost is purely debuggability: a client shipping a malformed proof can no longer tell which field drifted. Fine if intentional, just flagging it as a deliberate decision.

packages/linkstr/src/push/codec.ts
```diff
@@ -0,0 +1,85 @@
...
+      return yield* Either.left<PushWrapFailure>("invalid-signature");
+    }
+    if (
+      !wrap.tags.some(
+        (tag) =>
+          tag[0] === LINKY_PUSH_MARKER_TAG &&
```
kaladivo: Small reuse/cleanup batch in this file: the marker check and p-tag scan re-implement `internal/nostrEvent.ts` helpers — this check is `tagValues(wrap.tags, LINKY_PUSH_MARKER_TAG).includes(LINKY_PUSH_MARKER_VALUE)` and the recipient scan is `tagValues(wrap.tags, "p")` (same `tag[1] !== undefined` handling; precedent in `inbox/authenticateWrap.ts` and `chat/codec.ts`). That would leave one copy of the tag-extraction rules instead of three. Also, just below, the second `recipient === undefined` guard after `recipients.length !== 1` is unreachable (and `noUncheckedIndexedAccess` isn't enabled in this repo, so it's not required by the type checker) — the two guards can collapse into one. Relatedly, `httpAuth/codec.ts` `singleTagValue` re-scans its already-filtered one-element list via `firstTagValue`; `values[0]?.[1] ?? null` is behavior-identical (keep the preceding filter — switching it to `tagValues` would change duplicate-rejection semantics for valueless tags).

kaladivo:
All review findings are now addressed on this branch (12 commits on top of the migration):

**Blockers**
- `#linky` wire filter removed — the subscription is plain `{kinds:[1059], since}` with the marker enforced client-side, and the test fake now rejects unindexed multi-letter tag filters the way strfry does, so this class of bug fails at unit level.
- Backfill race fixed: the seen-set records only authenticated **live** emissions and the service no longer durably records suppressed backfill, so a wrap first seen during a reconnect replay is still delivered when a live copy arrives (the test that pinned the losing behavior now pins the rescue).
- Observability: linkstr stays console-free but reports invalid push-marked wraps, relay attempt closes with reasons, per-relay EOSE, and fatal watcher failures through callbacks; `apps/push` logs all of them.

**Rest**
- 10-minute per-relay subscription refresh restored (detects relays that drop a REQ without closing the socket).
- Cross-relay `alreadyHaveEvent` dedupe wired through the transport (optional param; no change for other callers) plus a cheap marker peek before schema/signature work — unrelated firehose events are rejected by a tag scan again.
- Ownership-proof error contract restored: structural failures → 400 with human-readable messages, crypto/authz failures → 401; sig/id hex tightened at the schema level.
- Inert `PUSH_EVENT_DEDUPE_TTL_MS` knob removed end to end.
- Shared FIFO seen-set extracted to `internal/seenWrapIds.ts` with direct eviction tests; the two inbox loops keep their intentional divergence documented.
- New `relayWatcher.test.ts` pins the delivery-decision ordering (backfill never recorded, duplicates suppressed, pruning); new `codec.test.ts` covers all decode failure paths.
- `docs/architecture.md` and the linkstr README reconciled with the final behavior.

Verified: `bun run check-code` clean; full unit suite green (linkstr 264, push 13, core 42, linkstr-react 26, web-app 601). Note the `local-stack` e2e only runs on pushes to main, so it will gate the Vercel promotion after merge rather than on this PR.

PR #280, opened 2026-08-25
Make sharing a chat image export the actual photo, not a dead blob: URL

### LINKED ISSUE #279: Sharing image from chat should export the data not `blob:url`
Received an image in chat (bank payment confirmation). Now need to share this image with someone.

The three dots only offers "Reply" and "Copy", which I guess I could copy the image and try to paste it somewhere but I didn't know how to do that. Standard iOS thing is to use system share functionality (all apps do that so users are used to it). Holding the image offered that menu:
Image

Then I pressed "Share" (what I'm used to doing from other apps) but that was sharing a string `blob:https://app.linky.fit/...` which iOS/mac doesn't know how to open (well the url isn't public so it can't even open it and it definitely shouldn't be a public url because otherwise random people could see my image attachments).

Instead this should export the actual data and let the system (iOS) handle it.

jarvisonsmart:
Closes #279

## Problem

Receiving an image in chat (e.g. a bank payment confirmation) and trying to pass it on was a dead end:

- The message's three-dots menu only offered **Reply** and **Copy** — no share.
- Long-pressing the image (the standard iOS gesture) surfaced WebKit's native callout, whose **Share** exported the literal string `blob:https://app.linky.fit/...` — a pointer into the current tab's memory that no other app or person can open.

The app actually already had a correct file-based share (`navigator.share` with a real `File`), but it was only reachable through the fullscreen viewer's footer, so users never found it before hitting the two broken paths.

## What changed

- **Long-press on a chat image now opens the app's message menu** instead of the native callout (`-webkit-touch-callout: none` on chat/viewer images, and the image bubble no longer swallows the long-press gesture).
- **The message menu gains Share and Save for decrypted image messages**, wired to the same file-based helpers as the viewer. Share is hidden on devices without `navigator.share`, and a failed (non-cancelled) share falls back to a file download. Copy — which only produced the raw `linky:image:v1:...` payload string — is hidden for image messages.
- **Share/download helpers moved to `app/lib/privateImageFile.ts`**, shared by the menu and the viewer.
- **Exported filename is now `linky-image.<ext>`** with the extension derived from the blob type, replacing the hardcoded Czech `linky-obrazek.jpg`.
- **Inspector events**: `ChatImageShared` / `ChatImageSaved` / `ChatImageShareFailed` rows on `app.log`, linked to the message rumor id, with glossary entries.

## Verification

- New unit tests prove the share payload is a real `File` (`files:[...]`, never `url`/`text`), the filename/extension derivation, the cancel-vs-failure handling, and the download anchor naming.
- Component tests drive the real menu on a real (mock-decrypted) image message: menu shows **Share/Save** without **Copy**, clicking Share calls `navigator.share` with `linky-image.jpg`; text messages keep **Copy**.
- `bun run check-code` and the full unit suite (133 files, 633 tests) pass.

Not covered: manual on-device iOS verification of the share sheet — worth a quick check on a real phone before release.

### DIFF
--- a/apps/web-app/src/App.css
+++ b/apps/web-app/src/App.css
@@ -2208,6 +2208,15 @@ button.settings-sensitive-action.is-armed {
   background: rgba(15, 23, 42, 0.7);
 }
 
+/* No native long-press callout on chat images: it would share the blob: URL,
+   which is useless outside this tab; long-press opens the message menu instead. */
+.chat-private-image,
+.chat-image-viewer-image {
+  -webkit-touch-callout: none;
+  -webkit-user-select: none;
+  user-select: none;
+}
+
 .chat-private-image {
   display: block;
   width: min(260px, 70vw);
--- a/apps/web-app/src/app/lib/privateImageFile.test.ts
+++ b/apps/web-app/src/app/lib/privateImageFile.test.ts
@@ -1,0 +1,134 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { InspectorRow } from "../../devtools/inspector";
+import { reportInspectorRows } from "../../devtools/inspector";
+import {
+  downloadPrivateImageBlob,
+  isCancelledShareError,
+  sharePrivateImageBlob,
+} from "./privateImageFile";
+
+vi.mock("../../devtools/inspector", () => ({
+  reportInspectorRows: vi.fn(),
+}));
+
+vi.mock("../../devtools/inspector/inspectorEnabled", () => ({
+  getInspectorEmissionEnabled: () => true,
+}));
+
+const reportedRows = vi.mocked(reportInspectorRows);
+
+const share = vi.fn<(data?: ShareData) => Promise<void>>(async () => undefined);
+
+const lastReportedRow = (): InspectorRow | undefined =>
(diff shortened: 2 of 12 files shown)


jarvisonsmart:
> 🤖 [fable] on Dave's behalf.

An independent verification pass (gpt-5.6-sol driving a real Chromium against the dev build) confirmed the share payload is a correct `File` (no `url`/`text`), the callout-suppressing CSS lands in the served output, and all checks pass — but it also found two real bugs in the first commit, both now fixed in 8ad15de:

1. **Long-press click-through** — a touch browser can synthesize a click after the long-press menu opens (reproduced with Pixel 7 touch injection), which also opened the fullscreen viewer underneath the menu. The message now swallows that post-long-press click via a capture handler. The new regression test was verified to fail against the unfixed code.
2. **Decrypt loop on the bank offer page** — the page re-renders every second for its countdown, and the inline `onBlobChange` identity re-triggered the bubble's decrypt effect, re-downloading and re-decrypting the confirmation image once per second (reproduced: 5 image requests in ~1s). The callback now lives in a latest-ref outside the effect deps, with a test pinning one decrypt across identity-changing re-renders.

Full suite is green (637 tests) and `check-code` passes. Still not covered: share sheet behavior on a physical iOS device.

PR #283, opened 2026-08-26
Chat PDF attachments, add-to-group from messages, mint selection redesign (26.9.2)

hynek-jina:
## Summary

Bumps release to `26.9.2` with three user-facing changes.

### PDF attachments in chat
- Images and PDFs share one encrypted kind-15 file-message shape (AES-GCM, Blossom upload). PDFs ≤ 2 MB, sent byte-for-byte with a `name` tag; `width/height/fileName` on `PrivateImage` are now optional.
- First page rendered via `pdfjs-dist` (own chunk, excluded from SW precache), opens in the existing full-screen viewer with save/share. Falls back to a plain file card if rasterization fails.

### Add contacts from a message to a group
- Contacts mentioned in a message can be added to a group directly from the chat.
- Bank payment offers show a clear "Payout rejected" state; snapshot rows carry `rumorId` so confirmation images correlate on the offerer's side.

### Main mint selection redesign
- Mints stacked vertically, selected one tinted with ✓, `cashu.cz` gets a green "Recommended" badge. Removed the "Selected mint" heading and melt-to-main button (still in the token list).
- Fee block under the selected mint:
  - **Cashu payments** – `input_fee_ppk` from `/v1/keysets`, shown in sats assuming ~6 proofs (`100 ppk → ~1 sat`).
  - **Lightning top-up** – `0 sat` (protocol guarantee).
  - **Lightning payments** – from `fee_reserve` of a melt quote for a 10 000 sat probe invoice issued by another production mint. Skipped for test mints; emits `mint.lightningFeeProbe` inspector rows.
- Custom mint input accepts `cashu.cz` → `https://cashu.cz`.

## Testing
- `bun run check-code` clean, unit tests green (new tests for fee probe, fee parsing, MintsPage, PrivateFileBubble, BankPaymentOfferDetailPage, codec).
- Probe verified against production: cashu.cz returned `fee_reserve: 100` for a 10 000 sat invoice → `~1 %`.

## Notes
- The fee probe creates one unpaid mint quote on a third-party mint per mint per visit of the page; could be cached in `feesJson` with a TTL if needed.
- Deleted mint rows (`isDeleted`) are still ignored by `refreshMintInfo`, so their Cashu fee stays "Unknown" — pre-existing behavior.


jarvisonsmart:
> 🤖 [fable + gpt-5.6-sol] on Dave's behalf.

Multi-agent review: the PR was split into its three feature chunks (one per commit) and each got an independent blast-radius review, plus a dedicated pass on the `linkstr` package boundary. Findings below were verified against the actual code before being included.

## What changed

**Add contacts from a message to a group** (`b266a034c`) — extends the existing "Add all contacts" action: after inserting ≥2 unsaved npubs from a message, a group picker replaces the action (pick existing, create new, or dismiss), updating `groupName`/`groupNamesJson`. Threaded through `useAppShellComposition` → `buildPeopleRouteProps` → `ChatPage` → `ChatMessage`. Also carries the ~10k-line `bun.lock` diff — that part is serialization churn only (all 925 locked package records are identical to main).

**Chat PDF attachments** (`030d8c155`) — PDFs ≤2 MB reuse the kind-15 encrypted file message (AES-GCM + Blossom). In `linkstr`, `PrivateImage.width/height` become optional and `fileName` is added; the codec emits `dim` only when both dimensions exist and a new `name` tag for filenames. New `PrivateFileBubble` renders page 1 via `pdfjs-dist` with save/share. This commit also carries unrelated ride-alongs: the `BankPaymentOfferDetailPage` "Payout rejected" state + always-running countdown, `CashuTokenNewPage` auto-save-on-change, and a multiline change in `messageEditorDom`.

**Mint selection redesign** (`67fc9ce7d`) — vertical mint list, ✓ tint, hard-coded "Recommended" badge for `cashu.cz`, scheme-less custom URL input. Fee block for the selected mint: `input_fee_ppk` from `/v1/keysets` (shown as ~6-proof sat estimate), Lightning top-up shown as 0, and a Lightning payment fee **probe** that creates an unpaid 10 000 sat NUT-04 quote at another production mint and submits it to the selected mint's NUT-05 melt-quote endpoint. Display-only: `getMintFeePpk` and the probe result feed no payment execution path (melt/autoswap/top-up still compute their own quotes).

## Linkstr boundary check

The convention holds. The wire-level half of the PDF feature was correctly added to `linkstr` (`chat/domain.ts` + `chat/codec.ts`, with a round-trip test), and no new frontend code builds/parses Nostr events, discriminates kinds, or touches gift-wrap — the send path still goes `ImageMessageDraft` → outbox, receive still goes `WrapInbox` → dispatcher. The mint and groups chunks don't touch Nostr at all.

One gray area (pre-existing, but enlarged by this PR): the app-side compact codec `privateImageMessage.ts` (`linky:image:v1:` persisted in Evolu) mirrors `linkstr`'s `PrivateImage` field-by-field, so the schema now lives in **four hand-synced places** (linkstr `domain.ts`, linkstr codec tags, app payload interface/compact keys, `chatInbox.ts` mapping) — and `useSendChatMessage.ts:137-173` reconstructs the linkstr type by re-parsing the app's own serialization. Suggestion: let `linkstr` export the canonical string serialization for `PrivateImage` (pure codec, no Evolu dependency needed) and collapse the duplication.

## Issues

### Blocker
- **`bun.lock` is out of date — CI is already red.** `apps/web-app/package.json` adds `pdfjs-dist@^6.2.108` but the lockfile has no entry for it (0 occurrences), and all four workflows run `bun install --frozen-lockfile`, so `unit-tests` fails in ~5s. Regenerate and commit `bun.lock`.

### Major
- **Old clients silently drop incoming PDF messages.** New PDF rumors omit the `dim` tag; the pre-PR decoder requires it, so on old builds the wrap fails as `invalid-image` → `WrapDropped` — the message never renders and there's no hint one was sent. `docs/architecture.md:115` claims old clients "show an unparsed PDF message as its raw content", but that's only true for the Evolu own-device sync path, not the wire path. At minimum fix the doc and make the trade-off explicit; consider a fallback text rumor for the transition window.
- **Fee probe hits two production mints on every page mount** (`MintsPage.tsx:45`). Each visit creates a real unpaid NUT-04 quote at a third-party mint plus a NUT-05 melt quote at the selected mint; the result cache is component-local with no TTL/persistence, so failures and re-navigations re-probe. That's quote load on operators and lets the two mints correlate the user via invoice + timing. The PR notes acknowledge this — I'd make the `feesJson` TTL cache (or explicit refresh) a pre-merge item, not a follow-up. Related: neither POST has a timeout/abort (`lightningFeeProbe.ts:45`), so an unresponsive mint leaves `…` forever and requests outlive navigation.
- **PDF receive path has no resource limits.** Every PDF bubble decrypts and renders immediately on chat open (`PrivateFileBubble.tsx:57`) — unlike images, which are viewport-gated — and `privateImageMessage.ts:571` buffers the full response *before* checking the declared size. `pdfPreview.ts:52` bounds width but not height/pixel area, and renders up to 30 pages of peer-controlled geometry. A hostile or merely large attachment is a memory-exhaustion path. Lazy-load near viewport, enforce a byte cap before buffering, and add a per-page pixel budget.
- **Peer-controlled MIME/filename enables arbitrary-file export.** Routing trusts the claimed `application/pdf` (`privateImageMessage.ts:225`) without magic-byte checks, and `PrivateFileBubble.tsx:118` preserves the peer's filename (e.g. `invoice.exe`, `invoice.html`) on save/share even when preview fails. Validate PDF bytes and sanitize/force the `.pdf` extension.

### Minor
- **Group assignment clears pending state before mutations succeed** (`useContactsMessagingComposition.ts:2984`) — a validation/update failure (e.g. `groupNamesJson` 1000-char limit) dismisses the only retry UI and can leave contacts partially grouped. Clear only after all updates succeed.
- **Group picker vs. message gestures** (`ChatMessage.tsx:1067`) — the input/buttons sit inside a bubble with long-press-menu and swipe-to-reply handlers; pointer events bubble into both. Stop propagation at the picker or ignore interactive descendants.
- **Weak numeric validation in fee code** — `fee_reserve: null`/object/string coerces to `0` → false "~0 %" (`lightningFeeProbe.ts:67`); negative/fractional `input_fee_ppk` is accepted (`mintInfoHelpers.ts:252`). Also, the ~6-proof estimate uses only the cheapest *active* keyset while NUT-02 charges per actual proof's keyset (incl. inactive), so real fees can exceed the display.
- **Inspector gaps** — the new group-assignment mutation emits no event; `mint.lightningFeeProbe` has no glossary entry and empty `links` (quote IDs discarded); PDF save/share reuses `ChatImageSaved`/`ChatImageShared` whose glossary definitions mean images.
- **`messageEditorDom.ts` multiline change breaks caret math** — the serializer adds a synthetic newline before block nodes (`:22`) but the offset calculation (`:73`) doesn't count it, so caret restore drifts in multiline content.
- **`BankPaymentOfferDetailPage.tsx:521`** now starts a 1 s interval unconditionally in every state (the timed-phase guard was removed).
- Closed-app push notifications aren't PDF-aware (`sw.ts:322` returns null for every `ImageBody` → generic server fallback).
- Czech plural: `"Přidat {count} kontaktů"` is wrong for counts 2–4 (`cs.ts:592`).
- `linkstr` domain models `width`/`height` independently optional; the encoder silently drops a lone dimension while the app parser rejects it — model dims as both-or-neither.
- `privateImageMessage.test.ts:63` uses an `as Record<string, unknown>` cast, against the repo's no-cast rule.

## Recommendation

**Do not merge yet.** Before merging:

1. **Regenerate `bun.lock`** — CI is failing on `--frozen-lockfile` right now.
2. **Decide the cross-version story for PDFs** — old clients silently drop them; at minimum correct `docs/architecture.md:115` so the trade-off is documented accurately.
3. **Tame the fee probe** — TTL cache (or explicit refresh) instead of probe-per-mount, plus timeout/abort.
4. **Harden the PDF receive path** — byte cap before buffering, viewport-gated loading, pixel budget, filename sanitization.

The minor items are fine as fast-follows. Consider also splitting the ride-alongs (bank offer countdown/rejected state, CashuTokenNewPage auto-save, editor multiline) out of the PDF commit next time — they widen the rollback surface of an otherwise well-bounded feature. The linkstr layering itself is clean and the payment-execution paths are untouched by the mint fee display, so once the four items above land this is in good shape.

PR #341, opened 2026-09-08
feat(linkshu): settle topups from a NUT-17 push

gorrdy:
A topup used to learn its invoice was paid only on the next tick of a 5-second poll. Mints that advertise NUT-17 websockets for the quote's method and unit now push the settlement, and the wallet acts on it immediately.

## The poll stays at full speed

Slowing the poll down is the obvious saving and is exactly what must not happen:

- the mint pushes each state **once**, so a socket the OS tears down while the app is backgrounded loses that push for good — nothing re-subscribes, and NUT-17 replays state only on subscribe;
- `mintQuotePaid` never speaks for a quote that is already `ISSUED`, which is the state a resume after a lost mint response has to see.

So the subscription only ever shortens the wait, and nothing depends on it. It also cannot mint on its own: the claim re-checks the quote over HTTP under the counter lock, which a test caught by failing with `mint reported the settled quote as unpaid` when the fake pushed `PAID` while HTTP still said `UNPAID`.

## Shape

`packages/linkshu/src/internal/quoteSubscription.ts` holds both halves — the support check (per method and unit, an unloaded mint info reading as "no") and an `Effect.async` wrapper whose interruption cancels the subscription. `Topup` races it against the poll; a socket that fails falls silent and the poll decides. `LoadedWallet` gained the one `on.mintQuotePaid` method this consumes.

`QuoteStateChanged` grew an optional `via` (`poll` | `subscription`) so an exported inspector log says which watcher saw a state; the app row appends `(websocket)`. A mint that advertises NUT-17 but only ever produces `poll` rows is a socket worth investigating.

## Verification

- `bun run check-code` clean; `bun run test` 1514 tests pass, three of them new.
- The unit tests were mutation-checked: removing the race makes the push test fail rather than pass by accident.
- Verified by hand on a self-hosted build against a real mint. Before: `PAID` 16.5 s after `topup.start`, from the poll's fourth tick. After: `PAID (websocket)` at 12.3 s, and the claim completed 0.65 s later.
- The integration test asserts only that a real mint's advertised support parses into the answer the branch takes. The dev mint's FakeWallet marks a quote paid on the first HTTP poll, so no websocket delivers a settlement there and an end-to-end push test would be untrustworthy.

## Not addressed here

`resumePending` runs at startup and on the browser's `online` event, but not when the app returns to the foreground. A topup whose push was lost to a torn-down socket therefore waits for the next poll tick after the page unfreezes. That gap predates this change and is worth its own issue.


gorrdy:
TLDR, brutálně to zrychlí čekání na příchozí platbu. Je to defacto instantní.

kaladivo:
Díky - super improvement :). Merguju!

PR #348, opened 2026-09-09
feat(linkstr): let runLinkstr persist the inbox cursor

### LINKED ISSUE #333: linkstr: `runLinkstr` cannot persist the inbox cursor
## Problem

`runLinkstr` (`packages/linkstr/src/headless.ts`) accepts `outboxStore` but not `inboxCursorStore`, so every headless inbox run starts with an in-memory cursor. The service worker and any script opening `WrapInbox` through `runLinkstr` replay the full `since` window on each run instead of resuming from the last checkpoint. `linkstrServices` and the React runtime already take the option.

Documented in `packages/linkstr/docs/getting-started.md` ("Two ways to run") and `inbox.md` as a known limitation.

## Proposal

Add an optional `inboxCursorStore?: Layer.Layer<InboxCursorStore>` to `LinkstrHeadlessConfig` and pass it through to `linkstrServices`, then have `sw.ts` supply `InboxCursorStore.fromStringStorage` if it opens the inbox. Update the two guides.

PeterXMR:
Closes #333.

`runLinkstr` accepted `outboxStore` but not `inboxCursorStore`, so every headless inbox run started from an in-memory cursor and replayed the full `since` window on each run.

- Add optional `inboxCursorStore` to `LinkstrHeadlessConfig` and pass it through to `linkstrServices`.
- Test: a cursor persisted in a `StringStorage` stub makes the opened inbox subscribe from that cursor (minus the NIP-59 backdate margin) instead of the `since` fallback.
- Update `getting-started.md` ("Two ways to run") and `inbox.md` (cursor section) to drop the known-limitation notes.

`sw.ts` is untouched: it only calls `fetchWrapEvent` by id and never opens the inbox, so it has no cursor to resume, and service workers have no `localStorage` to back one anyway.

---

**Second commit, e2e pipeline:** the first CI run failed in `evolu-quota-recovery.spec.ts` with the source mint logging `Rate limit 20/minute exceeded`. nutshell 0.20.3 honours `MINT_RATE_LIMIT=FALSE` for its slowapi decorators but not for `limit_websocket` → `assert_limit` on the NUT-17 quote-subscription websocket, which hits the limiter storage directly. With three test browsers sharing the runner IP that path can trip the 20/minute default whenever timing lines up. The commit raises both `MINT_*_RATE_LIMIT_PER_MINUTE` values in `docker-compose.dev.yml` and records it in the CLAUDE.md gotcha. It is independent of the cursor-store change and can be split out if preferred.


jarvisonsmart:

Reviewed the full PR for correctness, simple improvements, package conventions, and documentation. The `inboxCursorStore` pass-through is clean and uses the existing composition correctly. No blocking code issue found.

Added d7d5d034:

- Replaced the one-off subscription stub with the shared `FakeRelay`. The regression now receives an authenticated message, verifies the saved checkpoint, opens a fresh runner from that checkpoint instead of its `since` fallback, and verifies both subscriptions close. It fails if the cursor-store pass-through is removed.
- Removed the redundant global mint rate cap. The transaction cap is necessary: in the actual Nutshell 0.20.3 image, `assert_limit` rejects call 21 even with `MINT_RATE_LIMIT=FALSE`; 100 calls pass with the raised transaction cap.
- Updated the missing `AGENTS.md` gotcha to explain that WebSocket exception. The PR already updates the relevant package guides correctly; no architectural change was needed.

Validation passed:

- `bun run check-code` and all 310 linkstr tests.
- Browser-loaded `runLinkstr` against the local Nostr relay: received text, persisted the cursor, reloaded, and confirmed the next wire subscription uses `stored cursor - 172800` despite a different fallback.
- Two mobile PWA browser contexts against local development services: messages in both directions, receiver reload with the persisted cursor, no duplicate bubbles, and another message delivered after restart.
- Resolved Compose configuration confirms both mints inherit the transaction limit.

The headless test exercises the new option directly. The two-user PWA run checks inbox integration for regressions; the PWA already supplied its cursor through the React runtime. The service worker fetches a wrap by ID and does not open an inbox, so it needs no cursor change. Browser verification used the development PWA, with traces and a screenshot retained locally.

PR #350, opened 2026-09-09
feat(linkshu): add and forget known mints explicitly

### LINKED ISSUE #334: linkshu: no explicit way to add or forget a known mint
## Problem

`Mints` (`packages/linkshu/src/mint/Mints.ts`) exposes `info` and `knownMints` only. A mint becomes "known" as a side effect of any successful wallet load (`linkshu.seenMints.<mint>` in the `KeyValueStore`), and there is no way to forget one. Callers who want to register a mint before it holds funds must call `info` for the side effect, and a mistyped or dead mint stays in the set forever, where `Restore` and `Validation` keep probing it.

`packages/linkshu/docs/mints.md` documents the workaround.

## Proposal

Add explicit `addKnownMint(mint)` and `removeKnownMint(mint)` to `Mints` (removal should refuse while the mint still has `accepted`, `reserved`, or `issued` rows), emit inspector events for both, and update `mints.md`.

PeterXMR:
Closes #334.

`Mints` exposed only `info` and `knownMints`: registering a mint before it held funds meant calling `info` for its side effect, and a mistyped or dead mint stayed in the seen set forever for `Restore` and `Validation` to probe.

- `addKnownMint(mint)`: records the `linkshu.seenMints.<mint>` entry with no network round trip; idempotent.
- `removeKnownMint(mint)`: drops that entry. Fails with the new tagged `MintInUse { mint, rowCount }` while any stored row still names the mint.
- Both emit `mints.addKnownMint` / `mints.removeKnownMint` inspector operations.
- `mints.md` and `errors.md` updated; `describeTaggedCashuError` in the web app gets the new tag.

One deliberate deviation from the issue text: it suggested refusing removal only for `accepted`/`reserved`/`issued` rows. But `knownMints` is the union of seen mints and the mints named by stored rows in **any** state, so forgetting the seen entry while an `error` or `externalized` row still names the mint would leave the mint in `knownMints` and the call would silently do nothing. Refusing while any row names the mint keeps `removeKnownMint` honest; callers clear those rows first (`Tokens.deleteSpent` for `error` rows) and retry. Happy to switch to the narrower rule if you'd rather keep `knownMints` as-is and accept the no-op.


jarvisonsmart:

Reviewed the full PR, package conventions, documentation, and inspector integration. The package methods are concise and reuse the existing mint decoder and operation reporting. Keeping the removal guard across all token states matches `knownMints`; no package refactor was warranted.

Committed `3140251`:

- Corrected the docs: `Validation` checks stored rows, not the seen-mint set. Explicit restore candidates also bypass that set, including those supplied by the PWA.
- Clarified that spending or changing a token's state can leave its stored row behind. The guide now names the cleanup operations and their limits.
- Added the missing inspector explanations for `mints.addKnownMint` and `mints.removeKnownMint`.

Validation passed on the local dev PWA at port 5350 with the local mint and relay stack:

- `bun run check-code`.
- 20 mint/cache tests and 12 app error/inspector bridge tests.
- Browser-loaded public API: duplicate add, duplicate removal, and removal of an unknown mint; local mint info; a real 32-sat top-up; `MintInUse` with `rowCount: 1`, preserving the 32-sat balance.
- Both new explanations rendered in the inspector, including the failed removal event.
- Separate PWA wallet flow: top-up from 0 to 64 sat and reload preserving 64 sat.

The new methods have no PWA UI caller yet, so their runtime checks used the actual browser-loaded package with a disposable wallet and the app's inspector bridge. The PWA wallet checks used its normal UI and persistent storage. No two-user flow is affected by this PR.

Issue #261, opened 2026-08-15
Dev inspector endpoints: wildcard CORS, no auth, unredacted decrypted content


kaladivo:
Found during the audit of #241. New in that PR. **Dev-only**: the collector is a Vite plugin with `apply: "serve"` and only `configureServer`, so the endpoints exist only while a dev server runs on the developer machine — production builds contain no collector and `inspector.html` is not a build input.

`apps/web-app/server/inspectorCollector.ts` sets `Access-Control-Allow-Origin: *` on `/__inspector/events`, `/stream`, `/clear`, `/report`, with no token or origin allowlist. While `bun run dev` is running, **any web page open in the same browser** can fetch `http://localhost:5173/__inspector/events` cross-origin and read the recorded stream — which includes decrypted plaintext chat, recipient pubkeys, raw Cashu tokens, and private-image decryption keys, because operation `params` are the raw pre-encryption drafts forwarded without redaction. Pages can also POST forged rows or wipe the store.

Also: the append-only `.inspector/rows-<port>.ndjson` on disk holds the same payload (gitignored and truncated on restart, but uncapped, world-readable 0644).

Task: restrict CORS to the dev-server origin (or require a token), and consider redacting sensitive draft fields (token text, image keys) before they reach the collector/disk.

kaladivo:
> 🤖 [GPT-6] on Dave's behalf.

PR #329 addresses the reproduced cross-origin read/write exposure and restricts the dev-log directory and current file to their owner. It preserves same-origin inspector traffic and CLI access; six live HTTP tests cover the boundary.

Keeping this issue open for the remaining work: redact spendable token/private-image key material before every sink, then bound dev NDJSON retention. Decrypted message content is intentionally supported by the inspector, so the redaction policy should distinguish content from secrets. Verify the same protection across the viewer, persistent browser log, collector and exported captures.

PR #351, opened 2026-09-09
fix(linkstr): keep attachment encryption keys out of inspector rows

### LINKED ISSUE #336: linkstr: inspector rows leak attachment encryption keys from chat drafts
## Problem

The linkstr inspector is documented as never carrying key material, and identity keys are indeed never emitted. But `sendToPeer` (`packages/linkstr/src/internal/wrapSend.ts`) passes the draft verbatim as inspector `params`, and `Outbox` emits the normalized operation on `outbox.enqueue`. So `chat.sendImage` operations and queued `chat.image` rows carry `image.key` and `image.nonce` (the attachment AES-GCM key), and `InboxRouted.event` carries received `ImageBody` keys.

In development that is memory-only, but the persistent-log toggle at `#advanced/inspector` retains rows in browser storage for up to 24 hours and lets the user download them. Anyone holding such a capture plus the Blossom URL can decrypt the attachment.

`packages/linkstr/docs/inspector.md` now states the intended rule and what a captured timeline actually contains.

## Proposal

Redact attachment keys at the emission points: strip `image.key`/`image.nonce` from chat drafts before they become `params` (a `redactDraft` on the chat codec, or a per-vertical `inspectParams` hook), and redact `ImageBody` in `InboxRouted`. Add a test that walks every emitted event for the image test fixture and asserts the key bytes are absent, then tighten the guide back to the rule.

PeterXMR:
Closes #336.

Identity keys never reached the inspector, but chat drafts did: `sendToPeer` passed the draft verbatim as `params`, the outbox echoed the normalized `chat.image` operation on `outbox.enqueue`, and `InboxRouted` / `inbox.fetchWrapEvent` carried a received `ImageBody`. Each held the attachment's AES-GCM `key` and `nonce`, so a persisted or downloaded timeline plus the Blossom URL could decrypt the attachment.

- `redactAttachmentKeys` (`src/internal/redactAttachmentKeys.ts`): walks a value and drops `key`/`nonce` from any object that also has `encryptionAlgorithm`; values with nothing to redact come back by identity, so unaffected rows are unchanged.
- Applied inside `inspectOperation` and `inspectPlainOperation` (params and result), the outbox enqueue row, and both inbox routing emissions. That covers every present emission point and any future vertical using the shared taps.
- `inspectorEmission.test.ts` sends an image through the outbox, receives an incoming image wrap on the inbox subscription, and fetches it by id, then asserts every emitted event still names the image URL but never contains the key or nonce bytes.
- `inspector.md` states the rule (no key material, attachment keys included) instead of documenting the leak.

Note for #261: the same helper is the natural place to redact before the dev collector and persisted NDJSON log, but that is app-side and left for that issue.


jarvisonsmart:

Reviewed the full PR, package conventions, emission call sites, and documentation. One quick fix is in a3d2bca9: the shared operation helpers were traversing/redacting parameters even with the inspector disabled. Redaction now runs inside the existing lazy event builders on both success and failure. This also keeps redaction failures inside the inspector's error boundary. The regression test observed four unwanted parameter reads before the fix and zero afterward.

The package remains platform-independent, uses no type casts, and needs no additional abstraction. The inspector guide now explicitly documents lazy redaction. I found no other actionable issues in the reviewed paths.

Validation on this commit:

- `bun run check-code` passed.
- All 313 linkstr tests passed, including attachment redaction, send/outbox/inbox/fetch emission coverage, and the new disabled-inspector regression.
- Ran the PWA from this branch on local port 5351 against the local relay/Evolu stack. Two separate browser accounts exchanged an image and PDF, decrypted/rendered both, received seen receipts, saved files, and shared the matching decrypted bytes. The existing `private-attachments.spec.ts` passed in 21.3 seconds. Blossom storage and the native share endpoint were stubbed; encryption, relay delivery, decryption, and the UI ran normally.
- Inspected 63 captured dev-inspector rows from both accounts. All eight attachment objects across enqueue, send, and inbox events omitted `key`/`nonce` while preserving URL, hash, and correlation links. Receiver screenshots show both attachments rendered correctly.

PR #353, opened 2026-09-09
fix(linkstr): deliver payment telemetry on its own outbox lane

### LINKED ISSUE #265: Outbox: failing telemetry job can head-of-line block chat sends
Tradeoff introduced by 046bc8e (payment telemetry delivered through the linkstr outbox, #241).

The outbox worker is a single strict-FIFO lane with invisible retry and backoff capped at 60s. Now that telemetry jobs share that lane with chat/reaction sends, a telemetry job that keeps failing (e.g. the analytics collector's relays are unreachable while the user's chat relays are fine) retries at ≤60s intervals and delays every chat/reaction job queued behind it. Previously telemetry had its own delivery lane, so its failures could not affect chat. `MAX_ITEMS_PER_FLUSH = 10` in `useAnonymousPaymentTelemetry` bounds how many telemetry jobs can pile up per flush, but does not remove the blocking.

Options: a low-priority lane in the outbox (telemetry only delivered when the main lane is idle), skip-ahead scheduling on repeated failure, or a per-operation-class max-attempts that fails telemetry jobs terminally instead of retrying forever.

PeterXMR:
Closes #265.

Since telemetry moved into the outbox (046bc8e) it shared the single strict-FIFO worker with chat and reaction sends. A report the collector's relays kept refusing retried at up to 60s intervals, and every chat job queued behind it waited. Before #241 telemetry had its own delivery path, so this could not happen.

Of the three options in the issue I took the lane, because it restores the old isolation exactly and loses nothing:

- Delivery now runs in two lanes: **foreground** (`chat.*`, `reaction`) and **background** (`paymentTelemetry`). Each is its own FIFO with its own wake queue and worker fiber; `enqueue` wakes the job's lane, the browser `online` event wakes both.
- Ordering and retry rules are unchanged within a lane. No attempt cap: a cap of a few attempts would drop reports during any real offline period, and a report that never lands costs one retry per 60s on an idle lane, which is what chat already does.
- Test: a telemetry job whose recipient the relays keep refusing, then a chat send behind it. The chat send completes at once; once the collector accepts, both telemetry jobs land in order.
- `outbox.md` (retry and ordering rules) and `payment-telemetry.md` updated.

The Outbox class docblock had one phrase ("strictly FIFO") that this change made wrong, so I amended that sentence.


jarvisonsmart:

Reviewed the complete PR and the outbox storage, runtime lifetime, retry/wake behavior, inspector emission, and package documentation. No blocking code issue found. The two-lane implementation is small and preserves the existing persistence and acknowledgement contract.

Follow-up commit `823bf8b0`:

- Extends the blocked-telemetry regression to prove reactions also deliver, after chat and in foreground FIFO order.
- Records the lane scheduling decision in `docs/architecture.md`.
- Removes an app comment that incorrectly said telemetry still shares chat's FIFO lane. The package guides already describe the new behavior correctly.

Validation passed:

- `bun run check-code`.
- All 310 linkstr tests across 35 files, including 29 outbox/store tests.
- Two local PWA accounts with service workers active, using the local Nostr relay and FakeWallet mint. A real 2-sat Cashu payment generated telemetry. The browser test rejected only collector-addressed wraps, then proved chat, edits, and reactions still reached the peer. A sender reload preserved state and another message delivered while telemetry remained blocked. Restoring collector acceptance and dispatching `online` drained both outboxes. Observed 9 rejected telemetry attempts and 4 forwarded reports; final balances were 47 sat and 1 sat after mint fees.

The first browser run raced a navigation before the online-dispatch check. The complete rerun passed in 27.4 seconds; no application change was needed.

PR #355, opened 2026-09-09
chore: route backup-import wallet rows through the linkshu token store

### LINKED ISSUE #314: Route backup-import cashuToken writes through the linkshu token store
## Problem

`apps/web-app/src/app/hooks/useAppDataTransfer.ts` (~line 284–312) restores wallet tokens from a backup file by writing `cashuToken` rows with a raw Evolu `upsert`, bypassing `@linky/linkshu`'s token store. It is the last remaining writer to the wallet table outside `evoluTokenStore.ts`.

This is hygiene, not a live bug: the import only inserts rows linkshu never wrote (it dedupes by token text and deterministic id), so the token store's write overlay has no entry to go stale — the new rows simply appear on the next reconciling `loadAll`, like any external Evolu write. Restoring rows byte-for-byte without mint validation is also the intended contract of an import; the next NUT-07 check reconciles reality.

But every extra writer to the wallet table is a place where the "linkshu owns cashu rows" invariant can quietly erode (the misplaced-lane healer removed in `993b84a` existed precisely because a second writer once put rows in the wrong lane). Two concrete wrinkles today:

- The import falls back to the default app owner when `cashuOwnerId` is null. Unreachable in practice for seed logins (import is a manual settings action, lane derivation completes in well under a second), but it is the same fallback shape that was just removed from the composition root.
- `buildImportedCashuTokenPayload` duplicates knowledge (sparse-payload convention, deterministic ids) that the token store already owns.

## Proposal

Add an import-shaped operation to the linkshu lifecycle (e.g. `cashuTokenLifecycle.importRow({ originalTokenText, tokenText, state, error })`) that writes through the token store — deterministic id, sparse payload, active-lane targeting, and overlay registration all come for free — and swap the import loop to it. Delete `buildImportedCashuTokenPayload` and the owner fallback.

Semantics stay byte-level restore: no receive/swap, no validation, state and error persisted as the backup claims.

PeterXMR:
Closes #314.

`useAppDataTransfer` restored `cashuToken` rows from a backup with a raw Evolu `upsert`, the last writer to the wallet table outside `evoluTokenStore.ts`. It duplicated the deterministic-id and sparse-payload conventions the store owns and fell back to the default app owner when the cashu lane was null.

**linkshu**
- `Tokens.importRow(draft: ImportRowDraft)` → `Effect<TokenRowId, TokenAlreadyKnown>`. Byte-level restore: no receive, swap, or mint check; `state` and `error` persisted as the backup claims (`error` kept only on `error` rows, matching the read boundary). Dedup by either token text, which also makes a backup imported twice a no-op.
- Inspector: `TokenLifecycleChanged` with `reason: "import"` and a `tokens.importRow` operation whose params/result carry state and row id only, never token text.
- `insertRowInState` takes an optional `error` so the import can persist one.
- `tokens.md` gains an `importRow` section; `errors.md` lists the operation.

**web app**
- `CashuTokenLifecycle.importRow` runs it through the linkshu runtime; `useCashuWalletComposition` now returns the lifecycle so the shell can reach it.
- The import loop decodes each backup row into an `ImportRowDraft` (state normalized by `normalizeCashuTokenState`, defaulting to `accepted`) and calls `importRow`. `buildImportedCashuTokenPayload`, the owner fallback, and the app-side dedupe sets are deleted; ids, sparse payloads, active-lane targeting, and overlay registration come from the store.
- Behavior notes: the import is now async, and until the wallet runtime is up (seed resolution, well under a second) token rows are skipped while contacts still import. Rows whose text is not a cashu token are skipped at decode time instead of being written and hidden on read, which is the same visible outcome as before.
- `docs/architecture.md` records that the adapter is the only writer to the `cashuToken` table.


jarvisonsmart:

Reviewed the full PR, including the linkshu lifecycle API, Evolu adapter, app wiring, tests, inspector events, and package documentation. The package implementation fits the existing structure; I found no useful refactor that would justify adding complexity.

Added commit e6f6418e to fix two backup edge cases:

- Export now preserves `originalTokenText` in the existing backup `rawToken` field. A received token's current proofs differ from its original encoding, so dropping that identity could break deduplication after restoring the backup.
- A backup containing wallet rows now asks the user to retry when the wallet runtime is still loading, before writing any contacts. Previously those wallet rows were silently skipped, potentially reporting a partial import as successful. The message is translated into Czech, English, and German.

Added regression tests for both cases, an inspector glossary entry for `tokens.importRow`, and architecture notes. Corrected the package guide: `Validation.checkAll` checks accepted/error rows; it does not settle imported reserved/pending operations.

Validation on this commit:

- `bun run check-code` passed. Four existing hook-dependency warnings remain outside the changed code.
- 43 linkshu token tests and 19 app backup/adapter tests passed. Both new regressions failed before the fix.
- Ran the PWA at `http://127.0.0.1:5355` against the local Docker services. A real 32-sat token was received as 31 sats, exported, restored into a separate seed-login profile, imported again without duplication, and exported with matching token/state/error bytes. Balance survived reload, and restore made no mint mutation requests.
- Held wallet seed resolution during startup, attempted a backup with a contact and an 8-sat token, and verified it wrote neither. Releasing initialization and retrying imported both successfully.
- Checked the import operation in the inspector, including its glossary explanation and link to the matching token lifecycle event.

No remaining blocking findings from this review.

PR #368, opened 2026-09-12
[PoC] fix(linkshu): walk the derivation tree to its end on counter collisions

gorrdy:
> **Proof of concept** — opened as a draft to show the diagnosis and one way to fix it. Please review the approach (shared gap limit, tree walk in collision recovery) before merging; the constants and the UX copy are open to change.

## What happened

On cashu.cz on 12 Sep 2026 one Linky wallet (iOS Safari, 26.9.12) hit `outputs already signed` for two operations in a row, five attempts each (`MAX_SWAP_ATTEMPTS`), right after a 30-batch NUT-09 restore; a third manual try succeeded. Each attempt's probe moved the counter by at most 100 slots, so the counter was ~1000 slots behind the tree. The mint and proxy were healthy: no rate limits, every 400 was a deterministic answer on duplicate proofs/outputs.

## Root cause

Two limits combined:

1. **Restore stops inside the tree.** `batchRestore` gave up after 300 unsigned positions. Operations reserve 64-slot output blocks and `quoteClaim` advances the counter by a block *before* every attempt, so a few failed attempts in a row leave a gap > 300 with live signatures behind it. `advanceCounterTo` never moves backwards, so restore could not fix a lagging counter either.
2. **Collision recovery probes one 100-slot window.** With a counter far behind, every bounded attempt advanced ≤ 100 slots and collided again.

## Change

- `internal/counters.ts`: shared `DERIVATION_GAP_LIMIT = 1000`, documented.
- `restore/internal/scan.ts`: `RESTORE_GAP_LIMIT` derives from it (300 → 1000).
- `internal/collisionRecovery.ts`: the probe uses `batchRestore` from the colliding counter to the tree's end and retries just past the last signed slot. Fallback bump unchanged.
- New `collisionRecovery.test.ts`; Receive/Send/Topup/Melt tests mock the probe via `batchRestore` (`probe`/`probeCalls`) while reclaim paths keep `restore`.
- Web app: new `cashuClaimedByRecipient` (cs/en/de). A `send` transfer whose proofs are spent reads "the recipient has already claimed this token" instead of "Token is invalid or already spent" in `useCashuTokenChecks` and `returnCashuTokenToWallet`.
- Docs: `docs/architecture.md`, `packages/linkshu/docs/receive.md`, `restore.md`.

## Trade-offs / open questions

- Every restore and every collision recovery now ends with 10 empty NUT-09 batches instead of 3 (~1.5 s extra on cashu.cz).
- This does not explain *why* the counter lagged in the first place (most likely the seed used from two storages, e.g. Safari tab vs. installed PWA on iOS, or repeated failed mint attempts). It makes recovery a single round instead of a retry loop.
- 1000 is a guess sized at 64 × 5 attempts × 2 blocks with headroom; happy to size it from `SWAP_OUTPUT_BLOCK`/`QUOTE_OUTPUT_BLOCK` instead.

## Verification

- `bun run check-code` passes.
- `@linky/linkshu`: 308/308 tests (4 new).
- web-app tests for `app/hooks/cashu`, `i18n`, `cashuStoredError` pass.
- Evidence: cashu.cz mint container log 12 Sep 15:34–16:45 UTC, nginx `access-cashu.cz.log` from 16:15 UTC.

https://claude.ai/code/session_014fA2kWTrghmSkgLhbo1GSu


gorrdy:
Parking this for now — the diagnosis stands (see description), but we will come back to the counter recovery with a broader fix later. Closing the PoC and deleting the branch.

PR #369, opened 2026-09-14
feat(web-app): return unclaimed issued tokens in bulk or after a wait

gorrdy:
PoC: not synced between devices; is it something we want to be in linky?

An issued send has no expiry of its own: it stays open until the recipient redeems it or the sender takes it back, one token at a time, from the token detail. With many tokens handed out as QR codes that never got scanned, this is a chore, and the funds sit outside the balance meanwhile.

- Tokens page: "Return unclaimed" runs one pass over every issued send (useReturnUnclaimedTokens). The mint is asked first so a redeemed token closes as claimed instead of being re-received; TokenAlreadySpent from the return counts as claimed too; the pass stops on MintUnreachable / CounterLockTimeout because the rest would fail the same way. The toast tallies returned / claimed / failed.
- Advanced > Return unclaimed tokens: off, 1 h, 6 h, 24 h, 3 days, 7 days. When set, the same pass runs every minute for sends older than the wait (no mint call while nothing qualifies). Opt-in and off by default, because returning a token invalidates a QR someone may still intend to scan; the token detail says when the return will happen.
- One `cashu` / `send.returnUnclaimed` inspector row per pass with the tally and the operation links of the receives it triggered.


kaladivo:
@hynek-jina this needs to be considered from the UX perspective. Might be more usefull to show and give user an option to reclaim these tokens if they had not been claimed in some time window, but not do it automatically.

PR #371, opened 2026-09-14
fix(cashu): never strand a sent token's sats; return unclaimed tokens in bulk

gorrdy:
## Why

A cashu.cz user issued a 15 261 sat token, then deleted it. Restore brought back nothing, yet the mint still holds all 12 blind signatures of that swap and none of the proofs is spent.

Deleting a token called `Tokens.forget`, which only closes the send as `done`; its proofs stay `handedOut`. From there nothing could bring the funds back:

- `Restore` skips every stored secret whatever its state, so the mint is never even asked about them (unchanged here, on purpose: restore must never invalidate tokens the user handed out).
- `Tokens.returnToWallet` refused `done` sends.
- The token page hid "Return to wallet" for closed sends.

## What

- **No delete without refund.** The token detail page offers a sent token only "Return to wallet"; delete stays for failed receives. Deleting a QR token that nobody scanned now brings the sats back instead of stranding them.
- **Return unclaimed tokens in bulk.** Tokens page, two-tap button "Return unclaimed to wallet (n)" (`useReturnUnclaimedCashuTokens`): one `returnToWallet` pass over every send whose proofs are still handed out — `issued`, `externalized`, or `done` (a delivered chat send, or a token deleted before this fix). A `pending` send is a delivery in flight and is never touched. `TokenAlreadySpent` counts as claimed (linkshu closes it as such), the pass stops at the first `MintUnreachable`/`CounterLockTimeout`, the toast tallies returned / claimed / failed, and one `send.returnUnclaimed` inspector row records the pass. The arm step warns that tokens sent in chat and not yet redeemed will be invalidated.
- **linkshu**: `returnToWallet` accepts a `done` send while any of its proofs is still `handedOut`/`externalized`. A claimed send (proofs `spent`) is still refused with `InvalidTransferTransition`. `forget` is unchanged.
- **docs**: tokens.md, concepts.md, architecture.md, CHANGELOG.

Closed sends are deliberately *not* listed among the transfer pills and restore is untouched, so a chat-sent token is only ever invalidated by an explicit, confirmed user action.

## Tests

- `Tokens.test.ts`: `returnToWallet` re-receives a `done` send with handed-out proofs; `done`/`returned` sends with spent proofs are still refused.
- `cashuTransfers.test.ts` (new): `isUnclaimedSend` over every status.
- `CashuTokenPage.test.tsx`: a sent token has return-to-wallet but no delete; a failed receive keeps delete.
- `bun run check-code` clean, linkshu suite 305 passed, web-app page tests pass.

Related: #369 (bulk + timed return of *issued* tokens). This PR's bulk action is the minimal manual version and also covers already-closed tokens; the two will conflict in `CashuTokensPage.tsx` and `useCashuWalletComposition.ts`.


kaladivo:
Closing in favor of #374

PR #378, opened 2026-09-15
feat(linkshu): slow the topup poll down as the quote ages

gorrdy:
A topup quote was polled every 5 s until it settled or expired, which on a mint that states no expiry means 24 h. Most quotes are never paid (34 of 41 on cashu.cz over 2026-09-14/15), so nearly all of the polling load came from invoices nobody was looking at any more.

The poll interval now follows the quote's age from the persisted `createdAt`: 5 s for the first five minutes, one minute until the quote is an hour old, then five minutes. A resumed old quote therefore polls slowly from the start. The NUT-17 subscription is unchanged and still reports a settlement at once; the poll stays a safety net for clients whose socket a proxy or CSP blocks.

Per never-paid quote and day this is ~340 polls instead of 17 280.


gorrdy:
Is this worth considering? The more users we have, the higher the load on the mint. Most users will go through the NUT-17 WS notifications anyway...

hynek-jina:
Sounds reasonable

PR #379, opened 2026-09-15
fix(web-app): store complete mint info and let its icon replace fallbacks

iWarpBTC:
Mint icons were mostly wrong for two independent reasons; this fixes both.

**Stored mint info was cut at 1000 characters.** The cap dates from the old Evolu `mintInfo` table (`NonEmptyString1000` columns). Mint info moved to localStorage in Jan 2026 and the table was deleted in Feb, but the cap stayed. A NUT-06 `/v1/info` response is normally longer than that, so the stored JSON was unparseable, the icon lookup returned null, and only mints with a hard-coded override showed an icon. Fix: store the whole payload; rows that already hold a cut blob are repaired on load and re-fetched.

**A cached fallback icon hid the real one for the whole session.** The icon cache remembered whichever URL loaded first per mint, so a favicon loaded before `/v1/info` arrived kept winning. Fix: remember only URLs that failed to load, keyed by URL, and derive the candidate chain on every render. A fresh info icon is tried as soon as it appears; a URL that failed is never requested again that session.

Tests cover a realistic 1381-character info payload being stored whole, repair of a cut blob, and the icon precedence rules.


kaladivo:
Currently in the middle of evolu rewrite. 

Will let you know once that is merged so you can rebase :).

kaladivo:

Rebased this branch onto current `main` and pushed the result as `319733bc`. GitHub now reports no merge conflicts.

The rebase preserves the mint metadata and icon fixes. I also removed the obsolete `onMintIconLoad` prop from `ContactCard.test.tsx` to match the updated component interface.

Validation on the rebased branch: all 27 targeted tests passed, workspace typechecking passed, and changed-file formatting passed. ESLint reported no errors and one hook-dependency warning in `ChatPage.tsx`.

PR #401, opened 2026-09-16
fix(bank-offer): settle from local authorship, not counterparty-controlled snapshot

gorrdy:
## Security reasoning

**Vulnerability.** A proxy-payment ("bank offer") could be **settled — releasing cashu from the wallet — based on fields inside a message the counterparty controls**, not on anything the settler actually authored. The offer message store keeps only the latest snapshot per contact, so a genuine offer awaiting settlement is **indistinguishable there from a forgery**.

**Attack.**
- **Fabrication (V2).** Any saved contact gift-wraps a `kind:24135` snapshot with `offererPublicKey` = the victim's own key and `status: "bank_paid"`. The victim's app files it as *their own* offer (`isOutgoing`/`isCreatedByMe` derived from the `offerer` field) and renders the owner view with a one-tap settle button and no confirmation. Tapping it runs `settleBankPaymentOffer`, which pays `offerInfo.amountSat` to the message's contact — i.e. the victim pays the sender for a bank payment they never received. (Confirmed end-to-end.)
- **Amount tampering (V3).** Even in a real flow, `settleBankPaymentOffer` read `amountSat` from the counterparty's `bank_paid` snapshot, so a real payer could inflate (or lower) the released amount up to the balance.
- **Replay / double-settle.** A second `bank_paid` after settlement could settle again.

**Root cause.** `decodeBankOfferRumor` takes `offerer` from the message body and never compares it to the authenticated wrap author (`rumor.pubkey`); ownership and the settled amount are read from the incoming snapshot; there is no settle idempotency; and `isOutgoing` was derived from the forgeable `offerer` field.

**Why the fix is sufficient.** Ownership now comes from a **persistent, unforgeable local record** (`bankPaymentOfferAuthored.ts`): when we author an `offered` snapshot we persist `offerId → amountSat` in `localStorage`. An attacker cannot write our `localStorage`, and `offered` is only ever authored by the offerer, so this is the trustworthy answer to "is this my offer, and for how much?". Settlement then: (1) settles only offers we recorded creating, (2) releases exactly the amount we offered and refuses any mismatch, (3) never settles the same `offerId` twice, and (4) confirms amount + recipient before releasing. The inbox marks a snapshot outgoing by authenticated authorship (`isSelfAuthored`), not the `offerer` field. Display (chat card, detail, preview) is unified on the amount we authored, a mismatched incoming amount is surfaced as an explicit warning, and the settle action is hidden everywhere on mismatch.

## What changed

- `apps/web-app/src/app/lib/bankPaymentOfferAuthored.ts` — new persistent authorship + settled-id store.
- `useBankPaymentOffers.ts` — record `offerId→amountSat` when we send an `offered` snapshot.
- `useCashuWalletComposition.ts` (`settleBankPaymentOffer`) — ownership + amount pin + idempotency + confirmation.
- `useLinkstrInboxSync.ts` — `isOutgoing: isSelfAuthored`.
- `BankPaymentOfferDetailPage.tsx` / `BankPaymentOfferViews.tsx` / `ChatPage.tsx` / `ChatMessage.tsx` — display the authored amount, mismatch warning, hide settle on mismatch.

## Verification

End-to-end on a self-host: a legitimate settle (offerer created the offer, payer marks `bank_paid` with the offered amount) goes through and confirms the correct amount; a tampered `bank_paid` (lower or higher) is refused, and the card/detail show the offered amount with the mismatch warning and no settle button. Full web-app suite green (808/808).

> ⚠️ Security-sensitive: money loss until deployed. Offers created before this change have no authorship record and cannot be settled afterwards (new offers unaffected).

https://claude.ai/code/session_01QMHVSWzPo3WL1Trpn3SMM1


gorrdy:
Superseded by `febd6035` ("fix(security): authenticate bank offer roles and settlement terms"), which is already on `main` and fixes the same V2/V3 issue more robustly at the protocol layer.

This PR mitigated the forged-settlement problem client-side, by recording local authorship (offer id + amount) in `localStorage` and refusing to settle on a mismatch. `febd6035` instead authenticates the terms themselves: `BankOfferAuthorization` treats only the offerer's authenticated snapshots as establishing terms (payer snapshots may only advance them), and `getBankOfferForSettlement` gates settlement on the offer being locally authored (`offererPublicKey === myPubkey`), `bank_paid`, and matching an authorized snapshot — so a counterparty can no longer forge the settled amount or role.

Since that landed, this branch conflicts in exactly the files `febd6035` touched (`useCashuWalletComposition.ts`, `useLinkstrInboxSync.ts`) and would only layer redundant client-side logic on top of the proper fix. Closing as superseded.

PR #403, opened 2026-09-16
fix(inspector): redact cashu tokens from inspector logs

gorrdy:
A cashu token is a bearer instrument: its proofs carry the mint secrets that let anyone holding the text spend the funds, so it is key material under the inspector's no-key-material rule. Sent token drafts (`chat.sendToken` params) and received token messages (`TokenBody.token`) were emitted verbatim, so the encoded token sat in the in-memory buffer, the 24h on-device log, and any exported `.ndjson`.

Broaden the single redaction chokepoint (renamed
`redactAttachmentKeys` -> `redactInspectorSecrets`, already applied by `inspectOperation`, `inspectPlainOperation`, the outbox enqueue row, and the inbox routing row) to replace a `token` field holding an encoded token (`cashuA`/`cashuB`) with a placeholder, alongside the existing attachment key/nonce stripping. linkshu rows already log only proof ids and amounts, and wire rows carry only encrypted gift-wrap, so no other plaintext-token path exists.

`inspectorEmission.test.ts` now sends, enqueues, receives, and fetches a cashu token end-to-end and asserts no emitted row contains the token text. The inspector doc guide and the adding-inspector-events skill rule are updated to call a cashu token out as key material.

## Security reasoning

- **Threat model:** the inspector's opt-in on-device log persists rows for 24h (~25 MiB IndexedDB) and can be exported as `.ndjson`. A cashu token is a bearer instrument — its proofs carry the mint secrets that let *any* holder spend it — so a token sitting in that log or an export is live money that anyone with the file can reclaim, including funds still in flight to a recipient.
- **Why it counts as key material:** the inspector's own no-key-material rule already forbids "mint secrets"; an encoded token embeds exactly those. Sent (`chat.sendToken`) and received (`TokenBody.token`) messages were emitted verbatim, so the rule was being violated in practice.
- **Fix placement:** all inspector emission funnels through one redactor (renamed `redactAttachmentKeys` → `redactInspectorSecrets`), so redacting the `token` field there covers the send draft, the received routed event, the outbox enqueue row, and the plain-op path in one place. Audited the other emitters: linkshu rows already log only proof ids/amounts ("the proofs never travel"), and wire rows carry only encrypted gift-wrap — no other plaintext-token path exists.
- **Verification:** an end-to-end emission test sends, enqueues, receives and fetches a real cashu token and asserts no emitted row contains the token text. The `adding-inspector-events` skill rule now names a cashu token as key material so this can't be reintroduced.

https://claude.ai/code/session_01QMHVSWzPo3WL1Trpn3SMM1


kaladivo:
Fixed in #424

gorrdy:
Reopened and rebased onto current `main` (`0756a385`, v26.9.17). The rebase was clean and all linkstr tests pass; no code changes were needed.

Reopening because this was closed as "Fixed in #424", but #424 did **not** fix this path — it redacted the **Evolu debug views** (`8987c0c0`, `evoluDebugValue.ts`), a different surface. The inspector redactor (`redactAttachmentKeys`) still strips only attachment `key`/`nonce` and never the `token` field, so cashu tokens are still emitted verbatim on `main`.

Confirmed live on `0756a385` from a real device inspector capture: 8 occurrences of a full plaintext cashu token (356 chars) in `InboxRouted.event.body.token`, 0 redaction placeholders — details and a redacted sample are in #427. Note the same build has `fa311f72` (the retry-loop fix) active and attachment key/nonce stripping working, so the emission and redactor both run; the redactor simply lacks a `token` branch, which this PR adds.

PR #431, opened 2026-09-18
feat(web-app): standing orders (recurring payments)

gorrdy:
## Summary

Standing orders ("trvalé příkazy"): a contact or a Lightning address is paid every N hours, days, weeks, or months from the device that created the order, while Linky is open on it.

- **Schedule** (`app/lib/recurringSchedule.ts`): due times are `anchor + n × interval` evaluated as a wall clock in the order's own time zone — a Jan 31 monthly anchor pays Feb 28, Mar 31, Apr 30 instead of drifting; daily/weekly orders keep their time across DST. Minimum interval 1 hour. Catch-up is fixed to *pay once and skip*: however many periods passed while Linky was closed, one payment goes out and the schedule moves to the first due time after now.
- **Storage**: new Evolu table `recurringPayment` in the transactions owner lane (counted toward its rotation threshold, shown on the Evolu debug page). Each run is an ordinary `transaction` row carrying `recurringPaymentId` / `recurringDueAtSec`.
- **Engine** (`useRecurringPaymentsScheduler`): ticks on launch, every minute, on visibility, and on `online`; the pure planner (`recurringPaymentTick.ts`) decides run / wait-for-funds / skip / interrupted per order, with a 24 h grace window capped at the next due time, 10-minute retries after failures, and a device binding (`executorDeviceId`, `linky.device_id.v1`) so only one device pays. A run is claimed on its row before money moves and rolled back on failure. Payments go through the existing contact-chat and Lightning-address paths (`fromQueue` / `recurringRun` option: no overlay, no navigation). Contact runs are preceded by the chat note "Trvalý příkaz: <title>".
- **UI**: `#wallet/recurring` list (transaction-style rows, plus FAB), `#wallet/recurring/new` form (recipient header when prefilled, pill selectors for recipient kind and interval unit, sat amount, first payment, optional run limit and note), `#wallet/recurring/<id>` detail (pay now, pause/resume, pay from this device, two-tap delete). Wallet page links to the list; both pay pages offer "Repeat regularly…" prefilled with recipient and amount; transactions that belong to an order carry a tappable "Standing order" pill.
- Inspector rows `recurring.*`; docs in `docs/architecture.md` ("Recurring payments") and README.

Known limitation, stated in the UI: there is no server-side scheduler, so payments only run while Linky is open in the foreground on the paying device. Fiat-denominated amounts, NUT-18 POST recipients, and native reminder notifications are follow-ups.

## Test plan

- [x] Unit: schedule math (DST gap/fold, month-end clamping, far-future estimate), order row validation, tick planner (grace window, retries, device binding, stale running mark), scheduler hook (claim/pay/roll back, on-demand run), display helpers — 955 web-app tests green.
- [x] E2E (`tests/recurring-payments.spec.ts`, added to `local-stack`): A funds, saves B, creates an already-due order via "Repeat regularly…"; the scheduler pays it on the next pass, B receives the chat note and the token, A's history pill opens the order, a second pass does not pay again.
- [x] `bun run check-code` clean.
- [x] Manually on a self-hosted build against cashu.cz: contact and Lightning-address orders, pay now, pause/resume, delete.


gorrdy:
Need revision from the UI/UX perspective

kaladivo:
@hynek-jina

hynek-jina:
Thanks @gorrdy . 
I have a few points I’d like us to address before we merge this — but feel free to push back on any of them.

* Notifications aren’t being sent. I’d imagine something like: “A scheduled payment is ready — it will be sent shortly,” so the user still has a chance to cancel it.

* There’s no in-app notification either — the payment happens in the background. The user should see the same confirmation as with any other payment.

* Why is the payment tied to a specific device? Why can’t it be paid by whichever device happens to be online at the time? With Evolu, this shouldn’t be a problem.

* The input doesn’t respect the selected unit. If I have CZK selected, I’d like to see everything in CZK (even though the amount is currently stored only in sats). We can add support for storing it in a different base unit later.

* The feature is too prominent. I’d suggest hiding it in the transaction history. At the top, there would be scheduled payments, with an option in the top right to create a new recurring payment. (I wouldn’t call it a “standing payment” — it sounds too bank-like.)

* Setting up a payment is quite complicated. For example, I’d make every payment recipient-based — a recipient could also be an LN address without a Nostr account. The number of payments could also be selected in a more interactive way.

* Labels aren’t linked to the payment. This is a separate feature that we should polish properly.

* Don’t offer this in the “Pay contact” flow — instead, perhaps allow an already completed payment to be turned into a recurring payment?

gorrdy:
agree, this makes sense

hynek-jina:
It looks better, but there are still quite a few unresolved issues.

* When someone sends me a message, I get a notification even when the app is closed. I should get a notification in the same way when I’m about to send a recurring payment. That’s actually the main purpose of the notification — to prompt me to open the app (if that’s required to send the payment).

* The automatic payment also sends a message saying “Recurring payment” — I’d remove that for now until we figure out labels and messages for recipients.

* I should also get a notification when a recurring payment has been sent.

* In the transaction history, I’d have two transaction sections — “Scheduled” and “History.” The same way we separate different types of contacts — “proxy payment,” “conversations,” “other contacts,” etc.

  * At the top of the transaction history, I’d therefore see recurring payments (I assume people won’t have dozens of them — if they do, then it would make sense to hide them).
  * These payments should be sorted by the date of the next payment.
  * The app should make sure I have enough funds for each recurring payment. If I don’t, it should be subtly indicated that the payment is at risk of failing.
  * I’d add a new payment using a floating Plus button — just like when adding a contact or token.

* “Skip the next payment” feels a bit confusing and, more importantly, like a secondary edge case → I’d remove it.

* The 5-minute timer for sending is also too prominent. I’d remove it. It’s enough that I was notified once; after that, I’d simply send the transaction.

gorrdy:
@/tmp/claude-1000/-home-jan-Github-linky/71787733-7218-45b6-adc5-e80d10d887b7/scratchpad/pr431-reply.md

hynek-jina:
> @/tmp/claude-1000/-home-jan-Github-linky/71787733-7218-45b6-adc5-e80d10d887b7/scratchpad/pr431-reply.md

@gorrdy ?

hynek-jina:
Overview of the changes in the three commits on top of merge commit `87290e50`:

**Merge `main` (`87290e50`)**

* Merged the current `main` into the PR branch. For files that `main` had meanwhile migrated to `@linky/linksync` (shell, wallet composition, Evolu schema, transactions page), the `main` version was kept; the recurring-payment parts were added in subsequent commits.

**`43d91603` PaidOverlay and confirmation dialog**

* PaidOverlay is shown in the same sheet as the payment confirmation; sending and receiving use the same layout: counterparty avatar (or a checkmark) with an animated arrow (orange up = sent, green down = received), small name, large “Sent” / “Received”, and a large amount.
* `showPaidOverlay(title, details)` carries the direction, amount, and contact; it is passed through all payment paths: contact payment via chat, LN address and invoice, NUT-18 POST, token receipt (including the sender from chat), top-up, and npub.cash claim.
* `PaymentConfirmDialog`: optional `closeOnBackdrop`, `confirmProgress` button fill, `layout="description-first"` (amount displayed below the description).
* Added `paidHeadlineSent/Received` translations, CSS animations (respecting `prefers-reduced-motion`), updates to top-up and contact-payment tests, and a note in `docs/architecture.md`.

| prepared recurring payment | sent modal | received modal |
|---|---|---|
| Screenshot 2026-09-22 at 13 30 43 | Screenshot 2026-09-22 at 13 31 09 | Screenshot 2026-09-22 at 13 31 21 |
- all three include animations.

**`75e91739` Recurring payments**

* Moved the `recurringPayment` table to `@linky/linksync` (scope `transactions`, `makeRecurringPaymentsRepository`), including the schema, ID, table scope, tests, and package documentation (`concepts.md`, new `recurring-payments.md`). The app no longer touches Evolu at all.
* Rewrote the scheduler on top of the repository: in the background, it automatically pays on the minute tick; when the app is visible, it shows a confirmation dialog with the avatar and name, amount below the name, the text “Weekly recurring payment is ready”, a “Pay” button that fills up over 10 seconds and then sends the payment, and “Cancel this payment” (skips the period). Clicking outside the dialog does nothing.
* Reliability: a missed payment is executed on the first pass after startup as soon as the balance is sufficient; the 24-hour window was removed, so the entire period is waited out; an interrupted run is checked against transaction history and, if no record is found, the due date is restored.
* The amount is stored in the entered unit (sats, or fiat in cents) and converted to sats only when the payment is made; the converted side is displayed with “~”; without an exchange rate, the payment waits.
* Editing an existing payment (pencil icon in the top bar, route `#wallet/recurring/<id>/edit`): the form only allows day/week/month, shows the first payment date and time (cannot be in the past) and the next payment; there is no limit on the number of payments; hourly/“other” settings and the `maxRuns`/`endAtSec` limits were removed.
* List: date uses the history format, with an interval pill; the form no longer has a second background; the “Recurring payment” button was removed from the contact profile; the countdown modal was replaced with the shared dialog.
* Added new inspector events (`recurring.confirmationShown`, `waitingForRates`, `updated`, reason `cancelled` for `skipped`, `recorded` for `interrupted`), cs/en/de translations, e2e spec (payment after one minute, editing, cancellation from countdown), README, `docs/architecture.md`, and CHANGELOG.



**`e504800f` Push reminders (requires push server deployment)**
- @kaladivo , can you please have a look? 🙏 
* `apps/push`: `POST /reminders` endpoint replacing the set of reminders for a given pubkey under ownership proof, `recurring_reminders` table, `ReminderDispatcher` (every 15 seconds, 1-hour TTL, discards reminders older than one hour), `recurring_reminder` payload for web and Android, `PUSH_MAX_REMINDERS_PER_PUBKEY` limit, tests, README, and `.env.example`.
* Client: `useRecurringReminderSync` sends reminders for “one minute before the next due date” of all non-paused payments only when the set changes (or after 12 hours), and only when push is registered; the service worker displays localized text and tapping it opens the transaction history; added the `recurring.remindersSynced` inspector event and push service documentation in `docs/architecture.md`.

PR #436, opened 2026-09-19
fix(linkshu): bind each keyset id to its mint to stop counter-reuse theft

gorrdy:
## Summary

A malicious mint that copies an honest mint's keyset id could make the wallet reuse that mint's NUT-13 derivation slots and, combined with the proof store's upsert-by-secret, overwrite the victim's proofs and collect them on the next payment. The token that triggers it is auto-accepted from a chat message, so the victim does nothing.

## Why this is needed

NUT-13 deterministic secrets derive from `(seed, keysetId, counter)` **only** — the mint URL is not an input. linkshu, however, keys its deterministic counters by `(mint, unit, keysetId)` and `boundKeysetId` only checks that a keyset id is hex. Nothing rejects a keyset id that already belongs to a different mint, even though NUT-02 tells wallets to check for exactly this ("two keysets with the same id are the same keyset; a mint must not reuse another mint's id").

That gap chains into fund loss through three facts that are individually reasonable but fatal together:

1. **Counter scope is per mint URL, derivation is not.** A mint the wallet has never seen starts its counter at slot 1. So a second mint URL presenting an honest mint's keyset id gets a *fresh* scope at slot 1 and derives the **same** secrets and blinding factors the honest mint already signed at its slot 1, 2, 3… The mechanism was confirmed against the real code (no funds moved):

   ```
   counter key (honest): linkshu.detCounter.https%3A%2F%2Fcashu.cz.sat.00ad268c4d1f5826
   counter key (evil)  : linkshu.detCounter.https%3A%2F%2Fevil.example.sat.00ad268c4d1f5826
   secret at slot 1 (mint = cashu.cz)     : 8c361f8f…9b601ddf
   secret at slot 1 (mint = evil.example) : 8c361f8f…9b601ddf
   identical secret? true   identical blinding factor? true
   ```

2. **`ProofStore.insert` upserts by secret.** The row id is a pure function of the secret (`cashuProofIdFor(secret)`), so a proof arriving with a secret the wallet already stores overwrites the existing row — including re-pointing its `mint` to the attacker's URL.

3. **`Restore` skips known secrets.** An overwritten row is never recovered, so the honest-mint proof is gone from the wallet's view.

**Reachability.** The attacker needs no interaction: a `cashuB` token naming the evil mint is auto-received from a chat DM (`useChatMessageEffects`), which contacts the attacker's `/v1/keys`, `/v1/keysets`, `/v1/info` — a transparent proxy of the victim's real mint. The evil mint answers the swap with the honest mint's `C_` for the victim's `B_`, so the victim unblinds valid honest-mint proofs that are stored against the evil URL and overwrite the real rows. On the next payment, mint selection spends foreign mints first, swaps those proofs at the evil mint, and the attacker redeems them at the honest mint.

Why fix it **here** and this way:

- `WalletInstances.get` is the single choke point every vertical (send, receive, melt, topup, autoswap) obtains a wallet through, and it already has the `KeyValueStore`. Blocking here blocks the impostor everywhere, before it can be used for any operation.
- The alternatives are worse. Changing the derivation to include the mint URL breaks NUT-13 restore compatibility. Making `ProofStore.insert` key on mint as well as secret breaks the cross-device dedup invariant that the id is a pure function of the secret. A transparent proxy is byte-for-byte indistinguishable from the real mint by its keys, so no key/DLEQ check can catch it. Binding the keyset id to the first mint URL that presents it is the one place the wallet can draw the line without touching money-path math or storage identity.

## The fix

`WalletInstances.get` now:

- binds a keyset id to the **first mint URL** that presents it (`linkshu.keysetMint.<id>`, trust-on-first-use), and
- rejects with `MintRejected`, on every load **including the cached path**, any mint whose active keyset id is already bound to a different mint URL.

A keyset id is a hash of the keyset's public keys (NUT-02), so two honest mints never share one.

## Compatibility

Backward compatible for the normal one-mint-one-URL case, with no migration:

- The `linkshu.keysetMint.*` store is new and starts empty, so on upgrade every existing mint binds to itself on its next load — nothing is rejected. Proofs, counters, seen-mints and operations are untouched.
- The binding keys by the same normalized `MintUrl` the deterministic counter scope already uses, so a single mint never conflicts with itself (trailing slash / case are normalized away before reaching linkshu, as today).

One deliberate behavior change: a single mint reachable under **two different URLs** (a mirror, or a dev `https://…` + `http://localhost` setup) has its second URL rejected after the first binds. This is inherent to the defense — a proxy is indistinguishable from a mirror — and that pattern was already unsafe (two counter scopes over one keyset collide at the mint). The fix turns a silent collision into a loud rejection. Local mints served under one canonical URL, and separate mints with distinct keys (e.g. the E2E `cashu-mint` / `cashu-mint-target`), are unaffected.

Defence-in-depth follow-ups the audit also lists (bounding collision-recovery jumps, only auto-accepting tokens from already-trusted mints) are left for separate PRs.

## Test plan

- [x] New `WalletInstances` tests: a keyset id binds to its mint; a second mint with the same keyset id is rejected; the rejection holds on the cached path; the same mint keeps using its own keyset id across reloads.
- [x] Full linkshu unit suite green (348 tests); `@linky/linkshu` typecheck clean; `bun run check-code` clean.
- [x] `docs/mints.md` and `docs/architecture.md` updated in the same commit.

Reported by an internal security audit (LNK-01).


kaladivo (changes requested):
1. **Existing holdings remain vulnerable.** [WalletInstances.ts:246](/Users/kaladivo/workspace/linky/packages/linkshu/src/mint/internal/WalletInstances.ts:246) binds only the currently selected keyset, without accounting for existing proofs or historical keysets. After an upgrade or device sync, an attacker can claim an existing honest keyset first. I reproduced acceptance of that mint, then an overwrite of the honest proof through the real receive flow with a mocked colliding swap response. Establish ownership from existing holdings/history before accepting new bindings.

2. **Concurrent loads can bypass the binding.** [WalletInstances.ts:69](/Users/kaladivo/workspace/linky/packages/linkshu/src/mint/internal/WalletInstances.ts:69) checks and writes in separate operations. Two callers can both read an empty binding and both succeed. This reproduced in 10/10 runs with a scheduling gap after storage reads. Claiming a keyset needs mutual exclusion across contexts.

3. **Recovery from inactive-only mints now crashes.** [WalletInstances.ts:246](/Users/kaladivo/workspace/linky/packages/linkshu/src/mint/internal/WalletInstances.ts:246) reads `wallet.keysetId`, whose cashu-ts getter throws when no active keyset exists. I reproduced this with a real cashu-ts wallet. Previously supported restore access now fails with an unhandled defect.

gorrdy:
Thanks for the thorough review — all three are real and now addressed (force-pushed onto `fix/keyset-mint-binding`).

**1. Existing holdings remain vulnerable.** Ownership is no longer first-load-wins against an empty store. Before binding, `enforceKeysetMintBinding` resolves the owner from existing holdings (a proof already stored under that keyset id in `ProofStore`) and only falls back to the stored binding, so an impostor loaded first after an upgrade or a fresh device sync is rejected because the wallet already holds honest proofs under that keyset. First-load only wins when there are neither holdings nor a stored binding (a brand-new keyset), which is a DoS on that mint, never fund loss. Added a test that rejects a mint whose keyset id is held under another mint with an empty binding store.

**2. Concurrent loads can bypass the binding.** The check-and-write now runs under a per-keyset lease (`withKeyLease`, the same primitive the deterministic counters use), with a lock-free fast path only when the keyset is already bound to the same mint. Two concurrent loads of the same unbound keyset serialize; the second re-reads under the lease and is rejected. Added a test that runs two colliding loads concurrently against a real lease store and asserts exactly one binds.

**3. Recovery from inactive-only mints crashes.** `wallet.keysetId` is now read defensively; when cashu-ts throws because there is no active keyset (a restore-only load), the binding is skipped rather than crashing the load. That is safe because no new outputs are derived without an active keyset, so there is no counter reuse to guard. Added a test loading a wallet whose `keysetId` getter throws.

`docs/mints.md` and `docs/architecture.md` are updated in the same commit; full linkshu suite green.

Note: I force-pushed to keep a single commit and to drop an attribution line; sorry if that made the review delta harder to follow. Happy to switch to follow-up commits if you prefer.

jarvisonsmart:
> 🤖 [Fable] on Dave's behalf.

All three review points are addressed in 9ca66e84, and the PR does what it claims. The one thing I'd push back on before merging is the trust-on-first-use binding store, which adds a permanent lockout vector the description understates.

**Verification of the three review points**

- **Existing holdings.** Resolved. Before binding, ownership comes from any stored proof under that keyset id (spent rows included) and only then from the stored binding. The check reads only the active keyset, but that is sufficient: cashu-ts binds the wallet once at `loadMint()` to the cheapest active keyset and linkshu never rebinds it, so no other keyset is ever used for derivation. Impersonating an inactive keyset cannot cause counter reuse.
- **Concurrent loads.** Resolved. The check-and-write runs under a per-keyset lease, with a lock-free fast path only when the keyset is already bound to the same mint. Bindings are never rewritten, so the unlocked fast-path read is safe. The new test uses the real in-memory lease store and asserts exactly one of two concurrent claims wins.
- **Inactive-only mints.** Resolved. cashu-ts 4.5.1 throws from the `keysetId` getter when no active keyset exists (the wallet is left unbound rather than failing the load), so the try/catch skip is correct: nothing derives without an active keyset.

Verified locally: linkshu vitest 351 passed, `bun run check-code` clean (the four eslint warnings are pre-existing in the web app). Docker E2E not run.

**Is the change reasonable and minimal?**

The logic is about 100 lines at the single choke point, which is the right place, and it stays out of derivation math and storage identity. Concerns:

- **The TOFU lockout is worse than "a DoS on that mint".** Auto-accept in `useChatMessageEffects.ts` runs on every incoming message with no contact gating. One DM carrying a token from a proxy of mint X binds X's active keyset to the attacker's URL, and this device can never load X again. The same happens if an honest mint rotates its keyset and the attacker's token arrives before the user's next load; if the user already holds proofs under the old keyset, those funds become unspendable from this device. There is no unbind path and no UI, and the store is device-local localStorage, so only clearing storage recovers. The description should state this plainly, and I'd want at least a recovery path before merging.
- **A simpler shape may exist.** The precondition for theft is "this wallet already derived under keyset K at mint A". Proof rows record that across devices, and the existing `linkshu.detCounter.<mint>.<unit>.<keyset>` keys record it on this device. Deriving ownership from those two sources only would drop the new `keysetMint` store and its lease, and would remove the lockout of never-used mints. The trade-off is a narrow race: two mints presenting one keyset loaded concurrently on a device with no proofs and no counters yet. Worth weighing against the permanent-lockout vector.
- **Minor.** The holdings resolver returns the first matching row, so if proofs exist under two mints for one keyset the winner depends on row order. The default no-op `resolveHoldingsOwner` parameter exists only for tests. The `Effect.tap` on the binding is duplicated on both load paths. The `mints.md` paragraph and the `architecture.md` sentence repeat each other almost verbatim.
- **The PR description is stale.** It still describes pure first-load-wins and does not mention holdings, the lease, or the inactive-keyset skip. Please update it to match the code.

kaladivo:
Napiš pls co řešíš a proč, normálně svymi slovy. Přidělám to tam, takhle jen běháme v kruzích od agenta k agentovi...

gorrdy:
Cashu odvozuje secrety jen ze seedu a keyset ID (URL mintu do toho nevstupuje) a keyset ID je hash veřejných klíčů, takže útočníkům mint může zkopírovat keyset poctivého mintu a inzerovat stejné keyset ID. Protože linkshu drží counter per (mint url, unit, keyset ID), attacker mint restartuje counter a jde znova od 1, tím reprodukuje už utracené secrets už utracené u poctivého minut - a protože ProofStore identifikuje řádek jen podle secret, kolize keyset id umožní přepis/záměnu mezi minty, a tím ztrátu funds.
  
Řešení: V enforceKeysetMintBinding (WalletInstances.ts) při každém načtení peněženky napevno svážu keyset ID s prvním mintem, který ho vlastní (útočník si keyset nezabere první). Když ten samý keyset zkusí použít jiný mint, skončí to MintRejected.

Cashu.me to dělá podobně, že pokud už ten keyset zná, tak mint se stejným keyset vůbec nepřidá, a tím tohle eliminuje.

